PR #4, opened 2022-06-23
Fix exchange retransmission on slow network

mfucci:
Message retransmission acks were randomly breaking the connection.
This only happened on slow local network or if one of the two devices is slow to respond.

The whole message exchange logic is not 100% correct and the code is hairy to follow, so this whole class should probably be rewritten in the future...


jdesai61 (commented):
The changes look ok so far, but I don't have access to the Matter specs, so hard for me to assess conformance.

PR #6, opened 2022-06-28
Fix MDNS server

mfucci:
1 _ start the server so it will answer queries
2 _ handle ANY queries
3 _ remove incorrect binding to MDNS broadcast IP
4 _ fix response flag to indicate that the server is authoritative


jdesai61:
Even though these changes don't fully fix the multiple interface issue, the changes are required to fix issue listed above

PR #7, opened 2022-07-11
Support multi interface

mfucci:
* Add support for multiple interface in the MDNS server
* Refactor MDNS server to isolate low level UDP APIs
* Deprecate Singleton class, so multiple Matter servers can be launched simultaneously


jdesai61:
This fixes the issue of me having to specify IP and now I am able to pair using Android CHIPTool or "chip-tool pairing ethernet 222 20202021 3840 192.168.1.28 5540" command where I specify the IP of the node-matter device.  

I assumed this fix would also make the device discoverable (able to pair using "chip-tool pairing onnetwork 222 20202021") so that I don't have to specify the IP of the device. I tried but it didn't work because MdnsServer doesn't have any records to serve until I connect to it. Any thoughts?

mfucci:
Yes, I did not reverse engineered and implemented this part of the protocol yet.
I think the logic should be:
* before commissioning, advertise _matterc._udp.local. on MDNS
* after commissioning, advertise _matter._udp.local. on MDNS (without the "c")

Maybe we can close your original bug and you can file a new one for this?
This will have to wait until I am back from vacations: I need to be able to compile chip-tool command line util before I can test on my side.

jdesai61:
Sure - thanks - Jatin

jdesai61:
Re-open for merge

PR #13, opened 2022-08-03
Handle wildcards in read requests

mfucci:
Add support for requesting all the attributes in one read


jdesai61:
This change seems good - although it doesn't completely complete HA commissioning flow.

jdesai61 (approved):
Looks good

PR #14, opened 2022-08-04
Split Matter logic from MDNS and add caching

mfucci:
This simplifies the overall logic by isolating Matter specific MDNS usage from generic MDNS server.
Caching computed records and UDP sockets improves the performances,


jdesai61 (approved):
I can't test this right now, but looks fine. This is a better way of handling MDNS. I will remove my pull request.

PR #15, opened 2022-08-07
Add discovery over MDNS for commission

mfucci:


jdesai61 (approved):
I have tested this change and it works (gets to same stage with Home Assistant as before) - seems like node-matter thinks commissioning is complete, but HA does not.  Regardless, the MDNS changes work fine.

PR #25, opened 2022-09-12
Partial implementation of attribute subscription

mfucci:
Missing some features but seems to work

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -65,12 +65,12 @@ class Main {
                 ))
             .addProtocolHandler(Protocol.INTERACTION_MODEL, new InteractionProtocol(new Device([
                 new Endpoint(0x00, DEVICE.ROOT, [
-                    new BasicCluster({ vendorName, vendorId, productName, productId }),
-                    new GeneralCommissioningCluster(),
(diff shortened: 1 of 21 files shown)


jdesai61:
My development VM is out of commission due to a QNAP NAS bug.  I tried o checkout the code on another machine and tried to build, but get following errors.  Not sure if this is an issue with my nodejs or the actual code.

tsc
src/interaction/cluster/GeneralCommissioningCluster.ts:54:22 - error TS7008: Member 'attributes' implicitly has an 'any' type.

54     private readonly attributes;
                        ~~~~~~~~~~

src/interaction/cluster/OnOffCluster.ts:16:22 - error TS7008: Member 'onOffAttribute' implicitly has an 'any' type.

16     private readonly onOffAttribute;
                        ~~~~~~~~~~~~~~

mfucci:
Updated this to "partial implementation of attribute subscription".

Can you approve this change and will continue adding the support of subscription in a follow up PR?
This PR and the next one have large code change and depend on each other, so it will be a lot of work to merge any change I am making here.

jdesai61 (approved):
Except for noted known issues, this change looks good

PR #31, opened 2022-09-27
Log fabric Id and node Id

mfucci:


mfucci:
@javierbarellano4v Good to merge down this one as well?

PR #33, opened 2022-09-30
Major refactoring

mfucci:
Clean up the code base, no functional changes:

- renaming for clarity and consistency
- refactor of the cluster servers to use the same cluster specs as the client
- move around classes to have cleaner boundaries between server / client, Matter specific logic / generic low level logic


mfucci:
@jdesai61 : any chance to do a quick review of this CL?
The subscription fix is built on top of it.
No feature changes, so no need to test it with HA: this is just code cleanup.

jdesai61 (approved):
Lots of total changes - but since subscription code that I have been testing is using this already, incremental changes are minimal, approving it here without testing more

PR #49, opened 2022-10-04
Returns undefined when no IP on the interface on Node implementation

mfucci:
On the previous fix, I wrote an unit test to ensure this fix, however the unit test is using the fake implementation of the network provider and I forgot to update the node implementation.


jdesai61 (approved):
Change looks good. This needs to merge into subscription code as it makes testing less cumbersome (node-matter doesn't crash on random MDNS query).

mfucci:
Merged with subscription fix code

PR #68, opened 2022-10-12
Update Cluster definition to official specs #1

Apollon77:
Hi @mfucci,

as already told here a very first stuff to update the existing cluster definition and make them complete compared to the official specifications.

Notes/Need support:
* I tried to add a way to define "default values" for Attributes. While this should work that way on a code level it is not working on type level because in my eyes we then also need an AttributeSpecOptional type - and only that allows to specify the defaultValue, but then allows the user to not provide the attribute at all
* Add optional Attributes (kind oof see last but without a default value).

Here I would need "Typing support". I'm not that Type expert atm ... but trying t get better - but your dark types magic is far bejond my limits right now :-)

relates https://github.com/mfucci/node-matter/discussions/45

PS: This will not build right now ... new BasicInfo fields needs to be added in Device.ts :-)

BTW: Other questions: Do we care about the events defined in the XMLs?

Ingo


src/Device.ts
```diff
@@ -61,7 +61,7 @@ class Main {
         );
 
         // We listen to the attribute update to trigger an action. This could also have been done in the method invokations in the server.
-        onOffClusterServer.attributes.on.addListener(0, on => commandExecutor(on ? "on" : "off")?.());
+        onOffClusterServer.attributes.on.addListener(0, on => console.log(`On/Off cluster: ${on ? "on" : "off"}`));
```
Apollon77, line 64: uups ... will remove that again later :-)

src/Device.ts
```diff
@@ -75,9 +75,13 @@ class Main {
                    new ClusterServer(BasicClusterSpec, { vendorName, vendorId, productName, productId }, {}),
                    new ClusterServer(GeneralCommissioningClusterSpec, {
                         breadcrumb: 0,
-                        commissioningInfo: {failSafeExpiryLengthSeconds: 60 /* 1mn */},
+                        commissioningInfo: {
+                          failSafeExpiryLengthSeconds: 60 /* 1mn */,
+                          maxCumulativeFailsafeSeconds: 60 * 60 /* 1h */,
```
Apollon77, line 80: no idea if that value makes sense? but by definition it is not optional

src/matter/cluster/BasicCluster.ts
```diff
@@ -4,17 +4,43 @@
...
+
+/**
+ * This cluster provides attributes and events for determining basic information about Nodes, which supports both
+ * Commissioning and operational determination of Node characteristics, such as Vendor ID, Product ID and serial number,
+ * which apply to the whole Node. Also allows setting user device information such as location.
+ */
 export const BasicClusterSpec = ClusterSpec(
     0x28,
     "Basic",
     {
```
mfucci, line 23: Per the specs, some attributes are read-only, some are writable with various ACL level.
Maybe you can add in a comment what is what: this will help when this will be refactored to add read / write control and ACLs

src/matter/cluster/BasicCluster.ts
```diff
@@ -4,17 +4,43 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { StringT, UnsignedIntT } from "../../codec/TlvObjectCodec";
+import { StringT, UnsignedIntT, BooleanT, Field, ObjectT } from "../../codec/TlvObjectCodec";
 import { AttributeSpec, ClusterSpec } from "./ClusterSpec";
 
+const CapabilityMinimaStructT = ObjectT({
```
mfucci, line 10: nit: "Struct" can be removed to be consistent with other type names
"T" is for template.

src/matter/cluster/DescriptorCluster.ts
```diff
@@ -7,14 +7,23 @@
 import { ArrayT, Field, ObjectT, UnsignedIntT } from "../../codec/TlvObjectCodec";
 import { AttributeSpec, ClusterSpec } from "./ClusterSpec";
 
+const DeviceTypeStructT = ObjectT({
```
mfucci, line 10: nit: same here, maybe Stuct is not necessary

src/matter/cluster/GeneralCommissioningCluster.ts
```diff
@@ -45,18 +46,26 @@ const SetRegulatoryConfigRequestT = ObjectT({
...
 
+/**
+ * This cluster is used to manage global aspects of the Commissioning flow.
+ */
 export const GeneralCommissioningClusterSpec = ClusterSpec(
     0x30,
     "General Commissioning",
     {
-        breadcrumb: AttributeSpec(0, UnsignedIntT),
+        breadcrumb: AttributeSpec(0, UnsignedIntT), // INT64U ??
```
mfucci, line 56: int64u can only be represented as bigint in js, so you can use UnsignedLongT type instead.

src/matter/cluster/GeneralCommissioningCluster.ts
```diff
@@ -45,18 +46,26 @@ const SetRegulatoryConfigRequestT = ObjectT({
...
         commissioningInfo: AttributeSpec(1, BasicCommissioningInfoT),
         regulatoryConfig: AttributeSpec(2, RegulatoryLocationTypeT),
         locationCapability: AttributeSpec(3, RegulatoryLocationTypeT),
+        supportsConcurrentConnections: AttributeSpec(4, BooleanT, true),
     },
     {
-        armFailSafe: CommandSpec(0, ArmFailSafeRequestT, 1, SuccessFailureReponseT),
-        updateRegulatoryConfig: CommandSpec(2, SetRegulatoryConfigRequestT, 3, SuccessFailureReponseT),
-        commissioningComplete: CommandSpec(4, NoArgumentsT, 5, SuccessFailureReponseT),
+        // Arm the persistent fail-safe timer with an expiry time of now + ExpiryLengthSeconds using device clock
```
mfucci, line 63: Those comments should be in the /** */ format.

And yes! Visual Studio magically displays those comments when accessing the method on the generated types!
I just tried.

So ClusterClient(interactionClient, 0, GeneralCommissioningClusterSpec).armFailSafe will display this comment...
And it is also displayed on the implementation of the Cluster on the device side.

Same for the comments on attributes, Visual Studio correctly displays them on generated types.

src/matter/interaction/InteractionProtocol.ts
```diff
@@ -32,8 +32,9 @@ export class ClusterServer<ClusterT extends ClusterSpec<any, any>> {
 
         // Create attributes
         for (const name in attributeDefs) {
-            const { id, template } = attributeDefs[name];
-            this.attributes[name as (keyof ClusterT["attributes"])] = new Attribute(id, name, template, attributesValues[name]);
+            const { id, template, defaultValueDef } = attributeDefs[name];
+            const defaultValue = attributesValues[name] !== undefined ? attributesValues[name] : defaultValueDef;
```
mfucci, line 36: attributesValues[name] cannot be undefined right, so the defaultValueDef will always be ignored...
As suggested in my previous comment, maybe you can ignore handling default values for now.

The correct way to handle them:

- AttributeValues should make optional attributes that have a default value
- you should check here for the value presence ( attributesValues.hasOwnProperty(name) ) and not if it is undefined since undefined can be a valid value (yes, undefined and not present are two different states in js...).

src/matter/cluster/BasicCluster.ts
```diff
@@ -4,17 +4,43 @@
...
         vendorId: AttributeSpec(2, UnsignedIntT),
         productName: AttributeSpec(3, StringT),
         productId: AttributeSpec(4, UnsignedIntT),
+        nodeLabel: AttributeSpec(5, StringT, ''),
+        location: AttributeSpec(6, StringT, 'XX'),
+        hardwareVersion: AttributeSpec(7, UnsignedIntT, 0),
+        hardwareVersionString: AttributeSpec(8, StringT),
+        softwareVersion: AttributeSpec(9, UnsignedIntT, 0),
+        softwareVersionString: AttributeSpec(10, StringT),
+        manufacturingDate: AttributeSpec(11, StringT ), // optional
```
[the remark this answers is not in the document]
Apollon77: Thank you for your thoughts. I agree that it makes sense to change the field "properties" to a object because it is more future proof, but leaving the id and type separate parameters. (so your first approach)

I also think we need an "Optional type" or we need to completely leave the optional fields aside because as soon as I add it here it has effects e.g. in "Device.ts" that the attribute needs to be defined because the attributes type struct has that value and a definition ... so it is not really "optional" anymore.
I would see two options for now: 
* Yes I add optional as "comment" and after the PR is merged you add the typing support and adjust all that in one run. (but that might need Device.ts gets some "unneeded" attributed just because needed
* You add optional typing into master and I rebase and then directly use it in the PR
* You add optional typing into my PR and we so have it together.

What do you think?

src/matter/cluster/BasicCluster.ts
```diff
@@ -4,17 +4,43 @@
...
     0x28,
     "Basic",
     {
+        dataModelRevision: AttributeSpec(0, UnsignedIntT),
         vendorName: AttributeSpec(1, StringT),
         vendorId: AttributeSpec(2, UnsignedIntT),
         productName: AttributeSpec(3, StringT),
         productId: AttributeSpec(4, UnsignedIntT),
+        nodeLabel: AttributeSpec(5, StringT, ''),
+        location: AttributeSpec(6, StringT, 'XX'),
```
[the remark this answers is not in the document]
Apollon77: From specs:
> A default value defined in the Default column is not meant to be the value used when the server returns to factory fresh settings. Specified conformance for data fields may be optional or change over time. A default value is defined to complete dependencies when the actual data field value is not present.
(and some more ... from Core Specs 7.17.4, page 378

In fact yes it is a "if device do not return it then controller can assume default value" ... so did I used the wrong place?

I can also remove my "default value code try" again and we stick with a commentfor now

src/matter/cluster/BasicCluster.ts
Apollon77: WIll do

src/matter/interaction/InteractionProtocol.ts
Apollon77: Default stuff removed for now ...

src/matter/cluster/BasicCluster.ts
Apollon77: Please check below ... that way or do you like it differently better?

src/matter/cluster/BasicCluster.ts
```diff
@@ -4,17 +4,52 @@
...
+        localConfigDisabled: AttributeSpec(16, BooleanT), /* default: false, optional: true, writable: true */
+        reachable: AttributeSpec(17, BooleanT), /* optional: true, default: true */
+        uniqueId: AttributeSpec(18, UnsignedIntT), /* optional: true, length: 32 */
+        capabilityMinima: AttributeSpec(19, CapabilityMinimaT), /* writable: false */
     },
     {},
 );
+
+/*
+Events:
```
Apollon77, line 49: @mfucci What we do with them? Should we also add a definition format for them

Apollon77: @mfucci Any feedbsck/idea here? Else I will add them in such comment style for now ...

Apollon77:
follow up PR after rebase is #72

PR #70, opened 2022-10-14
Add optional / writable attribute support

mfucci:
Added support for optional / writable attribute and conformanceValue.
(this was tougher than expected...)

I tried to use the syntax Attribute(1, StringT, { optional: true }) but I was not able to get it working. Maybe it is possible, but I don't know how...

So I introduce those new Attribute syntax:

1. Attribute(id, template, conformanceValue?): mandatory, read-only attribute with an optional conformanceValue. The conformanceValue is used only client-side for backward compatibility when an attribute now mandatory is not present on the server-side.
2. OptionalAttribute(id, template): optional, read-only attribute. No conformanceValue, since this is incompatible with optional.

- on the client-side, get[attributeName]() will return undefined is not present
- on the server-side, the optional attribute can be used by specifying an initialization value. Handler implementations that require some optional attributes must indicate them as now mandatory by using UseOptionalAttributes(cluster, { attributeName: true })
- WritableAttribute(id, template): mandatory, read / write attribute. No conformationValue for now since I am not sure what should happen if the client to to set an attribute not present
- writable & optional attribute are not supported (for now)


Apollon77:
So far that I understand it, it makes sense as you described it. Maybe use "one of the updated attributes" from my PR to also add at least one example per case ( one with value, one optionsl or such), that you can verfy it  code-vise and I have an example (and proof I got it right) too

mfucci:
Good suggestion! I did part of the BasicCluster and I had to add support for extra use cases.
Ready to review now!

Apollon77 (approved):
Looks good, makes sense to me and should give me enough examples to implement the rest of the clusters ...

I will just have a bit of fun rebasing my branch but geeat new basis for it

PR #71, opened 2022-10-15
Fix unit tests workflow

mfucci:
After this commit, unit tests will run on every merge


Apollon77:
hehe, yes makes sense, I already wondered why tests were not running on merge ;-))

Apollon77 (approved):

PR #72, opened 2022-10-15
Update Cluster definition to official specs #1 (re)

Apollon77:
Update of PR #68 after rebase

relates https://github.com/mfucci/node-matter/discussions/45

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -73,6 +73,7 @@ class Main {
             .addProtocolHandler(new InteractionProtocol()
                .addEndpoint(0x00, DEVICE.ROOT, [
                     new ClusterServer(BasicCluster, {
+                        dataModelRevision: 1,
                         vendorName,
                         vendorId,
                         productName,
@@ -84,12 +85,20 @@ class Main {
                         localConfigDisabled: false,
                         softwareVersion: 1,
                         softwareVersionString: "v1",
+                        capabilityMinima: {
+                            caseSessionsPerFabric: 100,
+                            subscriptionsPerFabric: 100,
+                        }
                     }, {}),
                     new ClusterServer(GeneralCommissioningCluster, {
-                        breadcrumb: 0,
-                        commissioningInfo: {failSafeExpiryLengthSeconds: 60 /* 1mn */},
+                        breadcrumb: BigInt(0),
+                        commissioningInfo: {
+                            failSafeExpiryLengthSeconds: 60 /* 1min */,
+                            maxCumulativeFailsafeSeconds: 60 * 60 /* 1h */,
+                        },
                         regulatoryConfig: RegulatoryLocationType.Indoor,
                         locationCapability: RegulatoryLocationType.IndoorOutdoor,
+                        supportsConcurrentConnections: true,
                     }, GeneralCommissioningClusterHandler),
                     new ClusterServer(OperationalCredentialsCluster, {},
                         OperationalCredentialsClusterHandler({
--- a/src/matter/cluster/BasicCluster.ts
+++ b/src/matter/cluster/BasicCluster.ts
@@ -4,8 +4,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { BooleanT, StringT, UnsignedIntT } from "../../codec/TlvObjectCodec";
+import { BooleanT, StringT, UnsignedIntT, ObjectT, Field } from "../../codec/TlvObjectCodec";
 import { Attribute, Cluster, OptionalAttribute, OptionalWritableAttribute, WritableAttribute } from "./Cluster";
+
+const CapabilityMinimaT = ObjectT({
+    caseSessionsPerFabric: Field(0, UnsignedIntT),
+    subscriptionsPerFabric: Field(1, UnsignedIntT),
+});
 
 /**
  * This cluster provides attributes and events for determining basic information about Nodes, which supports both
@@ -16,9 +21,10 @@ export const BasicCluster = Cluster(
     0x28,
     "Basic",
     {
-        vendorName: Attribute(1, StringT),
+        dataModelRevision: Attribute(0, UnsignedIntT),
+        vendorName: Attribute(1, StringT), /* length: 32 */
         vendorId: Attribute(2, UnsignedIntT),
-        productName: Attribute(3, StringT),
+        productName: Attribute(3, StringT), /* length: 32 */
         productId: Attribute(4, UnsignedIntT),
         nodeLabel: WritableAttribute(5, StringT, ""), /* maxLength: 32, writeAcl: manage */
         location: WritableAttribute(6, StringT, "XX"), /* length: 2, writeAcl: administer */
@@ -34,6 +40,16 @@ export const BasicCluster = Cluster(
         localConfigDisabled: OptionalWritableAttribute(16, BooleanT, false), /* writeAcl: manage */
         reachable: OptionalAttribute(17, BooleanT, true),
         uniqueId: OptionalAttribute(18, StringT), /* maxLength: 32 */
+        capabilityMinima: Attribute(19, CapabilityMinimaT),
     },
     {},
 );
+
+/*
+Events:
+* server:
+   * 0: StartUp,
+   * 1: ShutDown,
+   * 2: Leave
+   * 3: ReachableChanged
+*/
--- a/src/matter/cluster/GeneralCommissioningCluster.ts
+++ b/src/matter/cluster/GeneralCommissioningCluster.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Field, JsType, ObjectT, StringT, Template, UnsignedIntT } from "../../codec/TlvObjectCodec";
+import { Field, JsType, ObjectT, StringT, Template, UnsignedIntT, UnsignedLongT, BooleanT } from "../../codec/TlvObjectCodec";
 import { TlvType } from "../../codec/TlvCodec";
 import { Attribute, Cluster, Command, NoArgumentsT, OptionalAttribute, WritableAttribute } from "./Cluster";
 
@@ -26,37 +26,52 @@ const CommissioningErrorT = { tlvType: TlvType.UnsignedInt } as Template<Commiss
 
 const BasicCommissioningInfoT = ObjectT({
     failSafeExpiryLengthSeconds: Field(0, UnsignedIntT),
+    maxCumulativeFailsafeSeconds: Field(1, UnsignedIntT),
 });
 
-const SuccessFailureReponseT = ObjectT({
+const CommissioningSuccessFailureResponseT = ObjectT({
     errorCode: Field(0, CommissioningErrorT),
     debugText: Field(1, StringT),
 });
-export type SuccessFailureReponse = JsType<typeof SuccessFailureReponseT>;
+export type CommissioningSuccessFailureResponse = JsType<typeof CommissioningSuccessFailureResponseT>;
 
 const ArmFailSafeRequestT = ObjectT({
     expiryLengthSeconds: Field(0, UnsignedIntT),
-    breadcrumbStep: Field(1, UnsignedIntT),
+    breadcrumbStep: Field(1, UnsignedLongT),
 });
 
 const SetRegulatoryConfigRequestT = ObjectT({
     config: Field(0, RegulatoryLocationTypeT),
     countryCode: Field(1, StringT),
-    breadcrumbStep: Field(2, UnsignedIntT),
+    breadcrumbStep: Field(2, UnsignedLongT),
 });
 
+/**
+ * This cluster is used to manage global aspects of the Commissioning flow.
+ */
 export const GeneralCommissioningCluster = Cluster(
     0x30,
     "General Commissioning",
     {
-        breadcrumb: Attribute(0, UnsignedIntT),
+        breadcrumb: WritableAttribute(0, UnsignedLongT, BigInt(0)), /* readAcl: view, writeAcl: administer */
         commissioningInfo: Attribute(1, BasicCommissioningInfoT),
         regulatoryConfig: Attribute(2, RegulatoryLocationTypeT),
         locationCapability: Attribute(3, RegulatoryLocationTypeT),
+        supportsConcurrentConnections: Attribute(4, BooleanT, true),
     },
     {
-        armFailSafe: Command(0, ArmFailSafeRequestT, 1, SuccessFailureReponseT),
-        updateRegulatoryConfig: Command(2, SetRegulatoryConfigRequestT, 3, SuccessFailureReponseT),
-        commissioningComplete: Command(4, NoArgumentsT, 5, SuccessFailureReponseT),
+        /**
+         * Arm the persistent fail-safe timer with an expiry time of now + ExpiryLengthSeconds using device clock
+         */
+        armFailSafe: Command(0, ArmFailSafeRequestT, 1, CommissioningSuccessFailureResponseT),
+        /**
+         * Set the regulatory configuration to be used during commissioning
+         */
+        updateRegulatoryConfig: Command(2, SetRegulatoryConfigRequestT, 3, CommissioningSuccessFailureResponseT),
+        /**
+         * Signals the Server that the Client has successfully completed all steps of Commissioning/Reconfiguration
+         * needed during fail-safe period.
+         */
+        commissioningComplete: Command(4, NoArgumentsT, 5, CommissioningSuccessFailureResponseT),
     },
 )
--- a/src/matter/MatterController.ts
+++ b/src/matter/MatterController.ts
@@ -12,7 +12,7 @@ import { PaseClient } from "./session/secure/PaseClient";
 import { ClusterClient, InteractionClient } from "./interaction/InteractionClient";
 import { INTERACTION_PROTOCOL_ID } from "./interaction/InteractionProtocol";
 import { BasicCluster } from "./cluster/BasicCluster";
-import { CommissioningError, GeneralCommissioningCluster, RegulatoryLocationType, SuccessFailureReponse } from "./cluster/GeneralCommissioningCluster";
+import { CommissioningError, GeneralCommissioningCluster, RegulatoryLocationType, CommissioningSuccessFailureResponse } from "./cluster/GeneralCommissioningCluster";
 import { CertificateType, CertSigningRequestT, OperationalCredentialsCluster } from "./cluster/OperationalCredentialsCluster";
 import { Crypto } from "../crypto/Crypto";
 import { CertificateManager, jsToMatterDate, OperationalCertificateT, RootCertificateT } from "./certificate/CertificateManager";
@@ -63,7 +63,7 @@ export class MatterController {
 
         // Use the created secure session to do the commissioning
         let interactionClient = new InteractionClient(() => this.exchangeManager.initiateExchange(paseSecureSession, paseChannel, INTERACTION_PROTOCOL_ID));
-        
+
         // Get and display the product name (just for debugging)
         const basicClusterClient = ClusterClient(interactionClient, 0, BasicCluster);
         const productName = await basicClusterClient.getProductName();
@@ -71,8 +71,8 @@ export class MatterController {
 
         // Do the commissioning
         let generalCommissioningClusterClient = ClusterClient(interactionClient, 0, GeneralCommissioningCluster);
-        this.ensureSuccess(await generalCommissioningClusterClient.armFailSafe({ breadcrumbStep: 1, expiryLengthSeconds: 60 }));
-        this.ensureSuccess(await generalCommissioningClusterClient.updateRegulatoryConfig({ breadcrumbStep: 2, config: RegulatoryLocationType.IndoorOutdoor, countryCode: "US"}));
+        this.ensureSuccess(await generalCommissioningClusterClient.armFailSafe({ breadcrumbStep: BigInt(1), expiryLengthSeconds: 60 }));
+        this.ensureSuccess(await generalCommissioningClusterClient.updateRegulatoryConfig({ breadcrumbStep: BigInt(2), config: RegulatoryLocationType.IndoorOutdoor, countryCode: "US"}));
 
         const operationalCredentialsClusterClient = ClusterClient(interactionClient, 0, OperationalCredentialsCluster);
         const { certificate: deviceAttestation } = await operationalCredentialsClusterClient.requestCertChain({ type: CertificateType.DeviceAttestation });
@@ -80,12 +80,12 @@ export class MatterController {
         const { certificate: productAttestation } = await operationalCredentialsClusterClient.requestCertChain({ type: CertificateType.ProductAttestationIntermediate });
         // TODO: validate deviceAttestation and productAttestation
         const { elements: attestationElements, signature: attestationSignature } = await operationalCredentialsClusterClient.requestAttestation({ nonce: Crypto.getRandomData(16) });
-        // TODO: validate attestationSignature using device public key 
+        // TODO: validate attestationSignature using device public key
         const { elements: csrElements, signature: csrSignature } = await operationalCredentialsClusterClient.requestCertSigning({ nonce: Crypto.getRandomData(16) });
         // TOTO: validate csrSignature using device public key
         const { certSigningRequest } = TlvObjectCodec.decode(csrElements, CertSigningRequestT);
         const operationalPublicKey = CertificateManager.getPublicKeyFromCsr(certSigningRequest);
-        
+
         await operationalCredentialsClusterClient.addRootCert({ certificate: this.certificateManager.getRootCert() });
         const peerNodeId = BigInt(1);
         const peerOperationalCert = this.certificateManager.generateNoc(operationalPublicKey, FABRIC_ID, peerNodeId);
@@ -123,7 +123,7 @@ export class MatterController {
         return new InteractionClient(() => this.exchangeManager.initiateExchange(operationalSecureSession, operationalChannel, INTERACTION_PROTOCOL_ID));
     }
 
-    private ensureSuccess({ errorCode, debugText }: SuccessFailureReponse) {
+    private ensureSuccess({ errorCode, debugText }: CommissioningSuccessFailureResponse) {
         if (errorCode === CommissioningError.Ok) return;
         throw new Error(`Commission error: ${errorCode}, ${debugText}`);
     }
@@ -186,7 +186,7 @@ class RootCertificateManager {
         const signature = Crypto.sign(this.rootKeyPair.privateKey, CertificateManager.rootCertToAsn1(unsignedCertificate));
         return TlvObjectCodec.encode({ ...unsignedCertificate, signature }, RootCertificateT);
     }
-    
+
     generateNoc(publicKey: Buffer, fabricId: bigint, nodeId: bigint): Buffer {
         const certId = this.nextCertificateId++;
         const unsignedCertificate = {
--- a/src/matter/cluster/DescriptorCluster.ts
+++ b/src/matter/cluster/DescriptorCluster.ts
@@ -7,11 +7,20 @@
 import { ArrayT, Field, ObjectT, UnsignedIntT } from "../../codec/TlvObjectCodec";
 import { Attribute, Cluster } from "./Cluster";
 
+const DeviceTypeT = ObjectT({
+  type: Field(0, UnsignedIntT),
+  revision: Field(1, UnsignedIntT),
+});
+
+/**
+ * The Descriptor Cluster is meant to replace the support from the Zigbee Device Object (ZDO) for describing a node,
+ * its endpoints and clusters.
+ */
 export const DescriptorCluster = Cluster(
     0x1d,
     "Descriptor",
     {
-        deviceList: Attribute(0, ArrayT(ObjectT({ type: Field(0, UnsignedIntT), revision: Field(1, UnsignedIntT) }))),
+        deviceTypeList: Attribute(0, ArrayT(DeviceTypeT)),
         serverList: Attribute(1, ArrayT(UnsignedIntT)),
         clientList: Attribute(3, ArrayT(UnsignedIntT)),
         partsList: Attribute(4, ArrayT(UnsignedIntT)),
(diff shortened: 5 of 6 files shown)


src/matter/cluster/BasicCluster.ts
```diff
@@ -34,6 +40,16 @@ export const BasicCluster = Cluster(
         localConfigDisabled: OptionalWritableAttribute(16, BooleanT, false), /* writeAcl: manage */
         reachable: OptionalAttribute(17, BooleanT, true),
         uniqueId: OptionalAttribute(18, StringT), /* maxLength: 32 */
+        capabilityMinima: Attribute(19, CapabilityMinimaT),
     },
     {},
 );
+
+/*
+Events:
```
Apollon77, line 49: SHould I describe events like here?

src/matter/cluster/BasicCluster.ts
```diff
@@ -16,9 +21,10 @@ export const BasicCluster = Cluster(
     0x28,
     "Basic",
     {
-        vendorName: Attribute(1, StringT),
+        dataModelRevision: Attribute(0, UnsignedIntT),
+        vendorName: Attribute(1, StringT), /* length: 32 */
```
mfucci, line 25: I think this is maxLength and not length.

Matter specs uses "length" for both, but in some case, it is the max length, in other it is the mandatory length for the string.
It will be important to know the different when this will be used for data verification during parsing / unparsing.

src/matter/cluster/BasicCluster.ts
```diff
@@ -16,9 +21,10 @@ export const BasicCluster = Cluster(
...
         vendorId: Attribute(2, UnsignedIntT),
-        productName: Attribute(3, StringT),
+        productName: Attribute(3, StringT), /* length: 32 */
```
mfucci, line 27: Same here: maxLength

src/matter/cluster/BasicCluster.ts
mfucci: Not sure it is necessary at this point since there are not supported.

Or you can just copy / paste the full event specs from the XML file in the comment: I don't know how they will structured here and just the information below are not enough (optional, priority, type...)

src/matter/cluster/GeneralCommissioningCluster.ts
```diff
@@ -26,37 +26,52 @@ const CommissioningErrorT = { tlvType: TlvType.UnsignedInt } as Template<Commiss
...
 
+/**
+ * This cluster is used to manage global aspects of the Commissioning flow.
+ */
 export const GeneralCommissioningCluster = Cluster(
     0x30,
     "General Commissioning",
     {
-        breadcrumb: Attribute(0, UnsignedIntT),
+        breadcrumb: WritableAttribute(0, UnsignedLongT, BigInt(0)), /* readAcl: view, writeAcl: administer */
```
mfucci, line 56: I think readAcl: view is the default so probably no need to specify it here.
I haven't see yet attributes with a higher read privilege, but there are probably some.

src/matter/cluster/GeneralCommissioningCluster.ts
```diff
@@ -26,37 +26,52 @@ const CommissioningErrorT = { tlvType: TlvType.UnsignedInt } as Template<Commiss
...
     {
-        armFailSafe: Command(0, ArmFailSafeRequestT, 1, SuccessFailureReponseT),
-        updateRegulatoryConfig: Command(2, SetRegulatoryConfigRequestT, 3, SuccessFailureReponseT),
-        commissioningComplete: Command(4, NoArgumentsT, 5, SuccessFailureReponseT),
+        /**
+         * Arm the persistent fail-safe timer with an expiry time of now + ExpiryLengthSeconds using device clock
+         */
+        armFailSafe: Command(0, ArmFailSafeRequestT, 1, CommissioningSuccessFailureResponseT),
+        /**
+         * Set the regulatory configuration to be used during commissioning
```
mfucci, line 68: nit: same here, missing [s] and [.] to be consistent with other doc.

src/matter/cluster/GeneralCommissioningCluster.ts
```diff
@@ -26,37 +26,52 @@ const CommissioningErrorT = { tlvType: TlvType.UnsignedInt } as Template<Commiss
...
+         */
+        updateRegulatoryConfig: Command(2, SetRegulatoryConfigRequestT, 3, CommissioningSuccessFailureResponseT),
+        /**
+         * Signals the Server that the Client has successfully completed all steps of Commissioning/Reconfiguration
```
mfucci, line 72: nit: maybe can be shorten into
Informs that all steps of Commissioning/Reconfiguration needed during the fail-safe period have been completed.
All commands are for the client to tell something the server...
Yes, I know that was the original comment in Matter codebase...

mfucci (approved):

src/matter/cluster/BasicCluster.ts
Apollon77: Ok, will add XMLs. Good idea for later reference

src/matter/cluster/BasicCluster.ts
Apollon77: I still left the PR as "draft" becauser I also wanted to check against the PDF specs :-)

src/matter/cluster/GeneralCommissioningCluster.ts
Apollon77: Ok, I remove then

mfucci:
@Apollon77 : code Copyrights / IP / etc. With this PR, you will be the first contributor other than myself.
Have you deal with copyrights / IP questions on previous open source projects?

Current, I wrote all the code so I marked them with a copyright in my name.
I suggest:

- we create a node-matter Contributor License Agreement that basically says that we give our IP to "node-matter authors" and we sign it (probably by adding our name in a file committed in this repository)
- I modify all files to have them copyrighted to node-matter authors

If we don't do that:

- it will be very hard to update the license in the future if ever needed
- library users might be exposed to IPs / patents issues so might limit adoption

Apollon77:
For your Copyright question: Puuhhh, good one ;-)
I worked mainly on MIT projects so far, but Apache 2 should be kind of same-ish with some relevant additions. I also checked how the hass python server did it . It is also Apache-2 without any infos in code files.

For me personally it is as follows: If I contribute to a project then it is clear to me that I give away my IP onto the "owner" of this project I contribute to ... but "formalizing" this the way you described is the 100% transparent way. I would support that. The only question is when to add people to contributors? should that be done foreach PR/Code addition (even fixing a typo) or is it just for "significant changes" (however we define them then) ;-)
So yes lets do that, it is the "Most transparent way".
Additionally we migth should start a "CONTRIBUTING.md" in the repo to contain a small checklist for contributions - do not need to be fancy, but things like stick to coding style, add tests, "Make the code better if you see something weird" ... and in fact the reference to that Contributor license and maybe "withcontributing you agree to it" - and a PR template that includes a checkbox to confirm accepting the Agreement ... ;-)

Apollon77:
> For CONTRIBUTING.md : good idea. Any chance you can do a first pass on it?

sure, will add it to this PR

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+help is appreciated, from creating issues, to contributing documentation, fixing
+issues and adding new features.
+
+Please follow the best-practice contribution guidelines as mentioned below when
+submitting any changes.
+
+### Contributions License Agreement
+
+Before contributing, please read the [Contributions License Agreement](./CLA.md).
+Every contributor must sign this agreement before any code is merged into the
```
Apollon77, line 13: @mfucci Do you want to make it for "any code" or should I adjust it to "any significant/feature adding/potentially intellectual property relevant code"? (But then always needs to be check and defined).

Apollon77:
@mfucci Please have a look at https://github.com/mfucci/node-matter/pull/72/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055 ... what do you think?

src/matter/cluster/BasicCluster.ts
Apollon77: you were right ... I started to sync with specs ... some smaller diffs

src/matter/cluster/GeneralCommissioningCluster.ts
Apollon77: This is rather the good question ... leaving it the same as official documents or adjust ... :-) Will think about

Apollon77:
@mfucci next iteration on existing clusters :-) I fear we need something more , see my comments 
https://github.com/mfucci/node-matter/pull/72/commits/ddf6368ba681ba3a7c262dc0b1e82ed09454410a

src/Device.ts
```diff
@@ -100,7 +100,14 @@ class Main {
...
                     }, GeneralCommissioningClusterHandler),
-                    new ClusterServer(OperationalCredentialsCluster, {},
+                    new ClusterServer(OperationalCredentialsCluster, {
+                            nocs: [],
+                            fabrics: [],
+                            supportedFabrics: 254,
+                            commissionedFabrics: 0,
+                            trustedRootCertificates: [],
+                            currentFabricIndex: 0,
+                        },
```
Apollon77, line 110: No idea what we need to use as defaults here :-( but with compleing the spoecs we need to set it ... 

I need to test how chiptool reacts still, will do tomorrow

src/matter/cluster/BasicCluster.ts
```diff
@@ -8,8 +8,8 @@ import { BooleanT, StringT, UnsignedIntT, ObjectT, Field } from "../../codec/Tlv
 import { Attribute, Cluster, OptionalAttribute, OptionalWritableAttribute, WritableAttribute } from "./Cluster";
 
 const CapabilityMinimaT = ObjectT({
-    caseSessionsPerFabric: Field(0, UnsignedIntT),
-    subscriptionsPerFabric: Field(1, UnsignedIntT),
+    caseSessionsPerFabric: Field(0, UnsignedIntT), /* min:3, default: 3 */
```
Apollon77, line 11: Should we also do a "Field default value" Type definition? ;-)

src/matter/cluster/OperationalCredentialsCluster.ts
```diff
@@ -4,93 +4,162 @@
...
 export const enum CertificateType {
     DeviceAttestation = 1,
     ProductAttestationIntermediate = 2,
 }
 export const CertificateTypeT = { tlvType: TlvType.UnsignedInt } as Template<CertificateType>;
 
-export const RequestWithNonceT = ObjectT({
-    nonce: Field(0, ByteStringT),
+export const AttestationRequestT = ObjectT({
+    attestationNonce: Field(0, ByteStringT), /* length: 32 */
```
Apollon77, line 31: Another question: keep the propoerty names near to the specs or simplyify and use the "cluster or message type" as reference and so simplify - as you did with just calling it "nonce"? has all it's pro and cons ... I personally think staying near to specs makes it easier to map code later to specs

CONTRIBUTING.md
mfucci: I would say any code.
The main goal of the CLA is that if we want to change to license later, or even sell a commercial license, we don't have to reach out to every single contributor.
So even fixing a type change in a comment should require to sign the CLA. I had to do this to fix a comment in Matter codebase.

Apollon77: Good so text stays that way.

src/matter/cluster/GeneralCommissioningCluster.ts
```diff
@@ -55,8 +55,8 @@ export const GeneralCommissioningCluster = Cluster(
     {
         breadcrumb: WritableAttribute(0, UnsignedLongT, BigInt(0)), /* writeAcl: administer */
         commissioningInfo: Attribute(1, BasicCommissioningInfoT),
-        regulatoryConfig: Attribute(2, RegulatoryLocationTypeT),
-        locationCapability: Attribute(3, RegulatoryLocationTypeT),
+        regulatoryConfig: Attribute(2, RegulatoryLocationTypeT), /* default: value of locationCapability */
```
Apollon77, line 58: How we do stuff like this? Do we need a "constructor method" additionally to the commands for the Cluster Server logics?

src/matter/cluster/server/GeneralCommissioningServer.ts
```diff
@@ -12,9 +16,18 @@ export const GeneralCommissioningClusterHandler: ClusterServerHandlers<typeof Ge
...
+    setRegulatoryConfig: async ({request: {breadcrumbStep, newRegulatoryConfig, countryCode}, attributes: {breadcrumb, regulatoryConfig, locationCapability}}) => {
+        const locationCapabilityValue = locationCapability.get();
+        if (locationCapabilityValue !== RegulatoryLocationType.IndoorOutdoor && newRegulatoryConfig !== locationCapabilityValue) {
+            return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};
+        }
+        if (locationCapabilityValue === RegulatoryLocationType.IndoorOutdoor) {
+            regulatoryConfig.set(newRegulatoryConfig);
+        }
+
+        // TODO countryCode should be set for the BasicInformationCluster.location!
```
Apollon77, line 28: How can I access the other Clusters ? :-) (Just started to look into an easy use case to see if I understood it)

src/Device.ts
```diff
@@ -52,16 +52,25 @@ class Main {
 
         // Barebone implementation of the On/Off cluster
         const onOffClusterServer = new ClusterServer(OnOffCluster,
-            { on: false }, // Off by default
             {
-                on: async ({attributes: {on}}) => on.set(true),
-                off: async ({attributes: {on}}) => on.set(false),
-                toggle: async ({attributes: {on}}) => on.set(!on.get()),
+                onOff: false, // Off by default
+                globalSceneControl: true,
```
Apollon77, line 57: The current typing requires to also set values for all optional fields ... So each implementation needs at least to have something for these values, also if they would never use them. WOuld it make sense to also allow these attributes here to be "type wise" optional?

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,109 @@
...
+/*
+  <bitmap name="OnOffControl" type="BITMAP8">
+    <cluster code="0x0006"/>
+    <field name="AcceptOnlyWhenOn" mask="0x01"/>
+  </bitmap>
+*/
+
+const OffWithEffectRequestT = ObjectT({
+    effectId: Field(0, OnOffEffectIdentifierT),
+    effectVariant: Field(1, OnOffDelayedAllOffEffectVariantT), /* OR OnOffDyingLightEffectVariantT depending on first field */
```
Apollon77, line 45: ANy idea how to react for that? This is kind of a "combi field" and the allowed values and their meaning for effectVariant depends on effectId. Combine into one type and Do via "Naming" or whats best? ;-)

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,109 @@
...
+}
+const OnOffDelayedAllOffEffectVariantT = { tlvType: TlvType.UnsignedInt } as Template<OnOffDelayedAllOffEffectVariant>;
+
+export const enum OnOffDyingLightEffectVariant {
+    "20PercentDimUpIn_0p5SecondsThenFadeToOffIn_1Second" = 0,
+}
+const OnOffDyingLightEffectVariantT = { tlvType: TlvType.UnsignedInt } as Template<OnOffDyingLightEffectVariant>;
+
+/*
+  <bitmap name="OnOffControl" type="BITMAP8">
```
Apollon77, line 37: I woulkd need a type definition for the "bit map" fields. or do we do ENUMS and users need to use enum.A & enum.B or such to set multiple bits and use a number? or what do we do?

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,109 @@
...
+
+/**
+ * Attributes and commands for switching devices between 'On' and 'Off' states.
+ */
 export const OnOffCluster = Cluster(
     0x06,
     "On/Off",
     {
-        on: Attribute(0, BooleanT),
+        onOff: Attribute(0, BooleanT, false), /* reportable: true */
```
Apollon77, line 61: Do we need the "reportable" info?

From specs
> The Subscribe interaction supports all attribute data. The "reportable" quality is supported by other interactions that only require attribute data with this quality to support interval or change reporting.

I would leave as comment for now if you do not see a direct need

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,109 @@
...
+         * if it is in its ‘On’ state it SHALL enter its ‘Off’ state. On receipt of the Toggle command, if the value
+         * of the OnOff attribute is equal to FALSE and if the value of the OnTime attribute is equal to 0, the device
+         * SHALL set the OffWaitTime attribute to 0. If the value of the OnOff attribute is equal to TRUE, the OnTime
+         * attribute SHALL be set to 0.
+         */
         toggle: Command(2, NoArgumentsT, 2, NoResponseT),
+        /**
+         * The OffWithEffect command allows devices to be turned off using enhanced ways of fading.
+         */
+        offWithEffect: Command(0x40, OffWithEffectRequestT, 0x40, NoResponseT), /* optional: true */
```
Apollon77, line 91: Should we add an "OptionalCommand" which is then also optional for implementing the cluster logic?

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,109 @@
...
+         */
+        onWithTimedOff: Command(0x42, OnWithTimedOffRequestT, 0x42, NoResponseT), /* optional: true */
     },
 );
+
+/*
+Features:
+  <bitmap name="OnOffFeature" type="BITMAP32">
+    <cluster code="0x0006" />
+    <field name="Lighting" mask="0x01" />
```
Apollon77, line 109: In fact this is the "This cluster SHALL support the FeatureMap global attribute." topic ... so a cluster server instace should have some global Fields (ssee 7.13 in core specs)

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,109 @@
...
 export const OnOffCluster = Cluster(
     0x06,
     "On/Off",
     {
-        on: Attribute(0, BooleanT),
+        onOff: Attribute(0, BooleanT, false), /* reportable: true */
+        globalSceneControl: OptionalAttribute(0x4000, BooleanT, true),
+        onTime: OptionalWritableAttribute(0x4001, UnsignedIntT, 0),
+        offWaitTime: OptionalWritableAttribute(0x4002, UnsignedIntT, 0),
+        startUpOnOff: OptionalWritableAttribute(0x4003, OnOffStartUpOnOffT), /* nullable: true, writeAcl: manage */
```
Apollon77, line 65: We also need a "nullable" because else noone could set the value to "null" when using the Cluster and null is a valid value in this case: 

From specs:
> If the value is null, the OnOff attribute is set to its previous value.

src/Device.ts
```diff
@@ -58,21 +63,131 @@ class Main {
...
+                        onOff.set(false);
+                        onTime.set(0);
+                    } else {
+                        onOff.set(true);
+                        if (onTime.get() === 0) {
+                            offWaitTime.set(0);
+                        }
+                    }
+                },
+                offWithEffect: async ({request: { effectId, effectVariant}, attributes: { onOff/*, globalSceneControl*/ } }) => {
```
Apollon77, line 93: Ut seems that because of "non Writable Attribute" definition for "globalSceneControl" I can not get it as attributes here - at least my IDE and TS say it is not allowed.
But "Writeable" for attributes just mean that they can not be written by a request from "external" ... but they still needs to be available internally and also writeable

src/Device.ts
```diff
@@ -58,21 +63,131 @@ class Main {
...
+                    // TODO implement logic with globalSceneControl, but before:
+                    // 1. globalSceneControl is not possible to add to "attributes" above because not writable from extern!! ... but allowed to read and writable internally!!
+                    // 2. globalSceneControl controls if "the server SHALL store its settings in its global scene" ... we need to add that first
+                    /* TODO: Also discuss specs if this should be done also "delayed when setting off" or "already when starting the fade/dim timers??"
```
Apollon77, line 97: requires implementation of the scenes cluster... will put on my list.

src/matter/cluster/BasicCluster.ts
```diff
@@ -22,16 +22,16 @@ export const BasicCluster = Cluster(
     "Basic",
     {
         dataModelRevision: Attribute(0, UnsignedIntT),
-        vendorName: Attribute(1, StringT), /* length: 32 */
-        vendorId: Attribute(2, UnsignedIntT),
-        productName: Attribute(3, StringT), /* length: 32 */
+        vendorName: Attribute(1, StringT), /* maxLength: 32 */
+        vendorId: Attribute(2, UnsignedIntT), /* type: vendor-id */
```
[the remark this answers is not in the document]
Apollon77: https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/chip-types.xml#L54-L83

mfucci:
Hi Ingo, sorry for the slow reply / review: I was in Europe last week and just back so I should have more time now!

src/matter/cluster/server/GeneralCommissioningServer.ts
mfucci: You can't directly... (and this is by design to enforce good separation and the ability to unit test)
Adhoc methods can be added on the server that can be assessed through the session.

Or you can force the creation of the basic cluster first then pass the attribute you need when you construct this cluster:

type GeneralCommissioningClusterConfig = {
    deviceLocation: AttributeServer<string>,
};

export const GeneralCommissioningClusterHandler: (conf:GeneralCommissioningClusterConfig) => ClusterServerHandlers<typeof GeneralCommissioningCluster> = conf => {

Now you can access any require cluster configuration from inside the handlers while still keeping a good separation between clusters.

mfucci:
I would recommend to split this PR into one PR per cluster.

Some are nearly ready, other are still quite far and require additional feature from the framework and a lot of code change outside the cluster definition.

This will also speed up the review...

Apollon77:
@mfucci No problem ... my week was horrible in office too :-) I fyou ever are in Germany Karlsruhe tell and we meet :-))

Apollon77:
For this PR: if possible I would love to merge and iterate on it. The library is building, in fact this are adjustments of clusters that existed before alreayd roughly, nothing is breaking if we do as is this time and we have TODOs, so we can finish one by one afterwards with PRs ... else I come into merge/rebase hell ... 

For next CLuster implementations that are new it is easier to do one PR per Cluster.

Let me check your comments and then we sync again. Ok?

src/matter/cluster/GeneralCommissioningCluster.ts
Apollon77: I copied it from the Chip-Tool XML file out of the "description" tags, but Ok i will rephrase it

src/matter/cluster/BasicCluster.ts
Apollon77: Cool ... did not had that idea on my radar :-)

src/matter/cluster/BasicCluster.ts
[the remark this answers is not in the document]
Apollon77: Ok, I leave it as comments as it is right now ... Should I create an "issue" as task for that?

src/matter/cluster/OperationalCredentialsCluster.ts
Apollon77: The issue with such Spec references is that they might change (if they ad d a section with new number then we run out of sync very fast. In fact the only reference that makes sense is the obvious one "see "Operational Credentials Cluster" specification in Matter Core Specs (in fcat withthe 3 PDFs I'm sonstantly searching and switching because all stuff is in different places ;-))) But in fact this is the obvious reference. ... so no idea if refernce make sense if we keep the naming because then you can also search for type names or such in the PDFs and have the place

src/matter/cluster/GeneralCommissioningCluster.ts
Apollon77: Hehe ... Also needed to read it multiple times: In fact you can only cover that in code ... it reads enhanced:

The default value of the attribute "regulatoryConfig" is the value of the attribute locationCapability in this cluster :-)

Thats why I meaned that maybe a "constructor" method is helpful that allows a code like

```
if (regulatoryConfig.get() === undefined) {
    regulatoryConfig.set(locationCapability.get()); 
}
```

src/matter/cluster/server/GeneralCommissioningServer.ts
Apollon77: In fact I see several cases where a cluster needs to access other clusters. Like several clusters need to Access the Scene cluster (and then scene cluster needs access again to the others - at least I would do a "please store the device state in scene or restore is later").

The "session" feels the wrong place for it. So yes the idea with a "Conf" will work for some cases but might not for others (Scene) - unless we put this into "session" but session for me is even more global

src/Device.ts
Apollon77: Bildschirmfoto 2022-10-21 um 22 00 22

This is what my WebStorm marks
Bildschirmfoto 2022-10-21 um 22 01 28

Teh the tsc build tells this

Apollon77: So I can not compile the TS without the values

Apollon77: I could add a TODO and you check it after merge because in fact it is ok to have the values there was just wondering that I need to

src/matter/cluster/OnOffCluster.ts
Apollon77: >  Hopefully we won't see this one too often...

yes :-))

Apollon77: Works ... tried something comparable but in wrong place :-)

src/matter/cluster/OnOffCluster.ts
Apollon77: awesome! I add it as TODO in these places and we adjust after merge

src/matter/cluster/OnOffCluster.ts
Apollon77: added TODOs for it

src/matter/cluster/OnOffCluster.ts
Apollon77: In fact I would see this as another object with the  "global features" when defining the cluster. I assume that "device level" will not be enough (Imagine a device that has a socket and also a light ... so you will have two onOff clusters maybe with different "global features". And yes I assume we end up in such stuff (when you see Homekit devices and such then this will allll be mixed later on)

Apollon77: PS: And in fact we will need that because there are things for onoff that are only relevant when Lighing and not when no lighing ... so pot. there will be logic differences based on global features,

src/matter/cluster/OnOffCluster.ts
Apollon77: Ok, will tell when I stunmble over it in a more relevant usecase :-)

src/Device.ts
Apollon77: Works that way, thank you

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
+# Contributing
```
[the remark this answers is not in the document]
Apollon77: done, https://github.com/mfucci/node-matter/pull/75

src/matter/cluster/BasicInformationCluster.ts
```diff
@@ -0,0 +1,66 @@
+/**
```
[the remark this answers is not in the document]
Apollon77: I renamed it in my IDE ... normally IDE does that :-( Will see if I can somehow fix that

src/matter/cluster/OperationalCredentialsCluster.ts
```diff
@@ -4,93 +4,165 @@
...
-        requestCertSigning: Command(4, RequestWithNonceT, 5, CertSigningRequestResponseT),
-        addOperationalCert: Command(6, AddNocRequestT, 8, StatusResponseT),
+        /**
+         * Sender is requesting a certificate signing request (CSR) from the receiver.
+         */
+        requestCertSigning: Command(4, CertSigningRequestRequestT, 5, CertSigningRequestResponseT),
+        /**
+         * Sender is requesting to add the new node operational certificates.
+         */
+        addOperationalCert: Command(6, AddNocRequestT, 8, OperationalCertificateStatusResponseT),
```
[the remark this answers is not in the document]
Apollon77: Did

src/Device.ts
```diff
@@ -55,16 +60,135 @@ class Main {
 
         // Barebone implementation of the On/Off cluster
         const onOffClusterServer = new ClusterServer(OnOffCluster,
```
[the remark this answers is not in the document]
Apollon77: Ok I can remove the logic for now and we later add into an own PR ... then I add now the "new methods" as Promise.resolve() solutions and a TODO for OptionalCommands

Apollon77:
Ok, in the end you are right ... this PR got too messy and problematic ... ignore it ... I will do new ones. I hope I get all of that sorted out ... maybe I find time on the weekend

src/Device.ts
Apollon77: Just for info: UseOptionalAttributes worked when I used it on the Device.ts ... but when I started moving the methods in to a matter/cluster/server/OnOffServer.ts with a "OnOffClusterHandler" struct like theothers there then I did not found the place to use this above trick. how it would need to look like?

src/matter/cluster/GeneralCommissioningCluster.ts
Apollon77: And OnOff cluster would need add Kai kind of a „constructor“
Because it needs to add subscribes to two attribute values because there are timers that need to be adjusted when the value changes.

mfucci:
Karlsruhe: the closest I have been from it is in the thermal baths of Baden-baden :-)
My parents lives in Geneva and my mum grew up in Munich so I will let you know the next time I am in Europe!

Apollon77:
looking forward :-)
Reminds me that I was not in Baden Baden Thermal baths since ages ;-) Not children compatible .......

PR #73, opened 2022-10-16
Add a logger system

mfucci:


src/Controller.ts
```diff
@@ -13,8 +13,11 @@ import { NetworkNode } from "./net/node/NetworkNode";
 import { getIntParameter, getParameter } from "./util/CommandLine";
 import { singleton } from "./util/Singleton";
 import { MdnsScanner } from "./matter/mdns/MdnsScanner";
+import { Time } from "./time/Time";
+import { TimeNode } from "./time/TimeNode";
 
 Network.get = singleton(() => new NetworkNode());
+Time.get = singleton(() => new TimeNode());
```
Apollon77, line 20: For what this is needed?

src/Device.ts
```diff
@@ -39,6 +41,7 @@ const ProductIntermediateCertificate = Buffer.from("308201d43082017aa00302010202
...
 
 Network.get = singleton(() => new NetworkNode());
+Time.get = singleton(() => new TimeNode());
```
Apollon77, line 44: Same question here :-)

Apollon77 (commented):
Hey, just the two questions.

And: Do you think it makes sense to document the Loger somewhere and so the option to modify default loglevel?

src/Device.ts
mfucci: This is the other entry point which also needs to define which environment should be used.

src/Controller.ts
mfucci: This PR also starts the time abstraction and makes it depends of the environment (similar to Network):

- TimeFake for test, the time can be completely controlled by API
- TimeNode when running on node.js
- TimeWeb when running in a browser (when this will be implemented...)
- Time[SomethingElse] if more environment are supported, like Espruino or webviews in mobile apps

All env dependent code will be grouped in the future into one include like import ".env/NodeEnv" in the entry point before any other imports. Let me if you think that I should do this in this PR.

I can also have Node the default implementation for the time for now.

mfucci:
Logger doc added!

src/Controller.ts
Apollon77: I got that ... question was more - why here imported because unused ... The Time stuff is used in the logger mainly right now, correct?

I just noticed that it is in but unused :-)

Apollon77 (approved):

PR #74, opened 2022-10-17
Add CLA and update copyrights

mfucci:


Apollon77:
@mfucci I leave it up to you to merge, because I will sign then in "my first official PR" :-)

PR #77, opened 2022-10-21
Update Cluster definition to official specs #1 (without OnOffCluster)

Apollon77:
Ok, this PR incorporates the changes from #72 except CONTRINUTING.md and OnOff Cluster.

I added Todos as we discussed in #72
When this is merged I would do the OnOffCluster and we can have other PRs solving the Todos.

Thank you for your opinion

### DIFF
--- a/src/matter/cluster/server/GeneralCommissioningServer.ts
+++ b/src/matter/cluster/server/GeneralCommissioningServer.ts
@@ -20,14 +20,25 @@ export const GeneralCommissioningClusterHandler: ClusterServerHandlers<typeof Ge
 
     setRegulatoryConfig: async ({request: {breadcrumbStep, newRegulatoryConfig, countryCode}, attributes: {breadcrumb, regulatoryConfig, locationCapability}}) => {
         const locationCapabilityValue = locationCapability.get();
-        if (locationCapabilityValue !== RegulatoryLocationType.IndoorOutdoor && newRegulatoryConfig !== locationCapabilityValue) {
-            return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};
-        }
-        if (locationCapabilityValue === RegulatoryLocationType.IndoorOutdoor) {
-            regulatoryConfig.set(newRegulatoryConfig);
+
+        let validValues;
+        switch (locationCapabilityValue) {
+            case (RegulatoryLocationType.Outdoor):
+                validValues = [RegulatoryLocationType.Outdoor];
+                break;
+            case (RegulatoryLocationType.Indoor):
+                validValues = [RegulatoryLocationType.Indoor];
+                break;
+            case (RegulatoryLocationType.IndoorOutdoor):
+                validValues = [RegulatoryLocationType.Indoor, RegulatoryLocationType.Outdoor];
+                break;
+            default:
+                return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};
         }
 
-        // TODO countryCode should be set for the BasicInformationCluster.location!
+        regulatoryConfig.set(newRegulatoryConfig);
+
+        // TODO countryCode should be set for the BasicCluster.location!
 
         breadcrumb.set(breadcrumbStep);
         return SuccessResponse;
--- a/src/matter/cluster/server/GeneralCommissioningServer.ts
+++ b/src/matter/cluster/server/GeneralCommissioningServer.ts
@@ -1,8 +1,12 @@
 import { Logger } from "../../../log/Logger";
 import { MatterDevice } from "../../MatterDevice";
 import { SecureSession } from "../../session/SecureSession";
-import { CommissioningError, GeneralCommissioningCluster } from "../GeneralCommissioningCluster";
-import { ClusterServerHandlers, UseOptionalAttributes } from "./ClusterServer";
+import {
+    CommissioningError,
+    GeneralCommissioningCluster,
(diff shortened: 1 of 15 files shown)


Apollon77:
BTW: In my IDE I clearly see that BasicCluster is a rename ...no idea why GitHub web doesn't show it that way ...

Bildschirmfoto 2022-10-22 um 00 14 25

src/matter/cluster/server/GeneralCommissioningServer.ts
```diff
@@ -14,9 +18,18 @@ export const GeneralCommissioningClusterHandler: ClusterServerHandlers<typeof Ge
         return SuccessResponse;
     },
 
-    updateRegulatoryConfig: async ({request: {breadcrumbStep, config}, attributes: {breadcrumb, regulatoryConfig}}) => {
+    setRegulatoryConfig: async ({request: {breadcrumbStep, newRegulatoryConfig, countryCode}, attributes: {breadcrumb, regulatoryConfig, locationCapability}}) => {
+        const locationCapabilityValue = locationCapability.get();
+        if (locationCapabilityValue !== RegulatoryLocationType.IndoorOutdoor && newRegulatoryConfig !== locationCapabilityValue) {
```
mfucci, line 23: It is hard to follow when this is triggering.
Maybe better to make the checks explicit:
if ((locationCapabilityValue === RegulatoryLocationType.Outdoor && (newRegulatoryConfig !== RegulatoryLocationType.Outdoor))
   return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};
if ((locationCapabilityValue === RegulatoryLocationType.Indoor && (newRegulatoryConfig !== RegulatoryLocationType.Indoor))
   return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};
if ((locationCapabilityValue === RegulatoryLocationType.IndoorOutdoor && (newRegulatoryConfig !== RegulatoryLocationType.Indoor || newRegulatoryConfig !== RegulatoryLocationType.Outdoor))
   return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};

Alternatively:
let validValues;
switch (locationCapabilityValue) {
  case (RegulatoryLocationType.Outdoor):
      validValues = [RegulatoryLocationType.Outdoor];
      break;
   case (RegulatoryLocationType.Indoor):
      validValues = [RegulatoryLocationType.Indoor];
      break;
   case (RegulatoryLocationType.IndoorOutdoor):
      validValues = [RegulatoryLocationType.Indoor, RegulatoryLocationType.Outdoor];
      break;
   default:
     throw new Error(`Unsupported location type: ${locationCapabilityValue}`);
}

if (validValues.findIndex(newRegulatoryConfig) === -1) 
  return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};

src/matter/cluster/server/GeneralCommissioningServer.ts
```diff
@@ -14,9 +18,18 @@ export const GeneralCommissioningClusterHandler: ClusterServerHandlers<typeof Ge
...
+            return {errorCode: CommissioningError.ValueOutsideRange, debugText: "Invalid regulatory location"};
+        }
+        if (locationCapabilityValue === RegulatoryLocationType.IndoorOutdoor) {
```
mfucci, line 26: This if is not necessary since this is done internally by the attribute and might lead to issue if more RegulatoryLocationType are introduced in the future.

src/matter/cluster/server/OperationalCredentialsServer.ts
```diff
@@ -58,7 +64,25 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
 
         // TODO: create ACL with caseAdminNode
 
-        return {status: Status.Success};
+        return {status: OperationalCertStatus.Success};
+    },
+
+    updateOperationalCert: async ({ request: {operationalCert, intermediateCaCert, }, session}) => {
+        // TODO add logic
```
mfucci, line 71: throw new Error("Not implemented");

This will help debug if ever this case is triggered since otherwise it will fail later in the flow.
(same below)

mfucci (commented):
The fact the renaming is not correctly detected is still blocking to merge this PR...
This will create merge conflict that will potentially lose changes if the renaming is not correctly recorded (any modifications on the file with the old name won't be merged into the file with the new name).

Can you check how to make git understand the renaming?
Here is a test of renaming on my side: https://github.com/mfucci/node-matter/pull/81 where you can see that github understood the rename.

Here is how I did it:

- checkout main
- rename one file
- create a new branch test-rename
- commit the rename into that branch
- publish that branch

Apollon77:
For rename issue ... the only idea I have is that it is because of having one commit for rename (which is done right), but having another commit which changes the new file ... then it seems to loose that renaming thing ... whyever :-(
Only Idea I now have is to rename it back again and merge the PR and then do a PR with exactly "that single rename commit" as standalone PR. Only idea I have is to try that way

src/matter/cluster/server/GeneralCommissioningServer.ts
Apollon77: Ok, I chose the second option because in my eyes better to read then the "big if clauses". Good idea

mfucci (approved):

Apollon77:
Thank you, rename in followup PR done

PR #78, opened 2022-10-22
Abstract time related API so time can be faked in tests

mfucci:


Apollon77 (approved):
makes sense.

Apollon77:
Should I squash and merge or do you want to use normal merge?

mfucci:
In general, it might be better for reviewers to just approve the PR and let the contributor do the merging.
There is cleanup associate with merging: delete local branch / delete remote branch / merge the changes with the other branches in flight / etc...
What do you think? Maybe to add in Contributions.md ? Or we should make it clear in the file how the approval / merging is done?

Here, yes, merge without squash would be better

Apollon77:
> In general, it might be better for reviewers to just approve the PR and let the contributor do the merging.

Do you really want to give "all" contributors that send in a PR also write rights to the repo? No idea how you wanted to "Build up the project". I know it more that there in many cases will be a "core team of devs" that have write rights and many more contributors fropm outside.

I agree with you for "contributors with write access" - there this makes sense, for others obviously not :-)
I can add that to the Contributing file.

mfucci:
OK, makes sense.
Or we can also not overthink this: just for me don't merge since I would prefer not and we can leave this out of contributors since this is maybe too prescriptive for not that much value.
But I will also survive if you do merge my PR each time ;-)

Apollon77:
;-)
I think this rule "core team contributors" merge PRs fro others, but not from other core contributors" makes sense

PR #81, opened 2022-10-24
Test for renaming

mfucci:


Apollon77:
Try now to change the cntent of the renamed file please ... lets see if this is then same issue as with me

mfucci:
Now with content change in 2nd commit

Apollon77:
grmppffff ... no idea why it behaved that way with my Webstorm :-( Ok ... as said ... I do rename in extra PR

Apollon77:
Thank you for the test

PR #84, opened 2022-10-25
Add all device types based on 1.0 specs

Apollon77:
I added all device types based on specs. The Matter XML file of chip project contain 3 more that I could not find in specs. I left them out.

### DIFF
--- a/src/matter/common/DeviceTypes.ts
+++ b/src/matter/common/DeviceTypes.ts
@@ -5,6 +5,63 @@
  */
 
 export const DEVICE = {
+    // Utility Device Types
+    AGGREGATOR: { name: "MA-aggregator", code: 0x000e },
+    POWER_SOURCE: { name: "MA-powersource", code: 0x011 },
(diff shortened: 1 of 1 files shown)


src/matter/common/DeviceTypes.ts
```diff
@@ -5,6 +5,63 @@
  */
 
 export const DEVICE = {
```
mfucci, line 7: It would be good to indicate the reference to the specs next to each device, this will help when there is a need to go back the docs.

For instance for ON_OFF_LIGHT:
/** Lighting device that is capable of being switched on or off, see [Matter Device Library R1.0], section 4.1 */

Can be done in a follow-up PR.

mfucci (approved):

src/matter/common/DeviceTypes.ts
Apollon77: Will provide now ...

Apollon77: Re-Review please :-)

mfucci (approved):

PR #85, opened 2022-10-26
OnOff Cluster implementation

Apollon77:
As discussed here a "minimal/pseudo" implementation of OnOff Cluster that defines all attribues, commands and also shows a bit f the "topics" to discuss and decide.

thats why it is a lot of Todos and comments inside and yes it will not build right now ;-) But the main goal for now would be to show you better what I mean sop that we discuss "on topic" and not too theoretically. I hope it helps.

### DIFF
--- a/src/matter/cluster/OnOffCluster.ts
+++ b/src/matter/cluster/OnOffCluster.ts
@@ -8,34 +8,62 @@ import { BooleanT, Field, ObjectT, StringT, Template, UnsignedIntT, UnsignedLong
 import { Attribute, OptionalAttribute, OptionalWritableAttribute, Cluster, Command, NoArgumentsT, NoResponseT } from "./Cluster";
 import { TlvType } from "../../codec/TlvCodec";
 
-export const enum OnOffStartUpOnOff {
+
+// From [Matter Application Cluster Specification R1.0], section 1.5.5.1
+// If the value is null, the OnOff attribute is set to its previous value when the device starts up.
+export const enum StartUpOnOff {
+    /**
+     * Set the OnOff attribute to FALSE when the device starts up.
+     */
     Off = 0,
+    /**
+     * Set the OnOff attribute to TRUE when the device starts up.
+     */
     On = 1,
+    /**
+     * If the previous value of the OnOff attribute is equal to FALSE, set the OnOff
+     * attribute to TRUE. If the previous value of the OnOff attribute is equal to TRUE,
+     * set the OnOff attribute to FALSE (toggle).
+     */
     Toggle = 2,
 }
-const OnOffStartUpOnOffT = { tlvType: TlvType.UnsignedInt } as Template<OnOffStartUpOnOff>;
+const StartUpOnOffT = { tlvType: TlvType.UnsignedInt } as Template<StartUpOnOff>;
 
-export const enum OnOffEffectIdentifier {
+// From [Matter Application Cluster Specification R1.0], section 1.5.7.4.1
+export const enum EffectIdentifier {
     DelayedAllOff = 0,
     DyingLight = 1,
 }
-const OnOffEffectIdentifierT = { tlvType: TlvType.UnsignedInt } as Template<OnOffEffectIdentifier>;
+const EffectIdentifierT = { tlvType: TlvType.UnsignedInt } as Template<EffectIdentifier>;
 
-export const enum OnOffDelayedAllOffEffectVariant {
-    FadeToOffIn_0p8Seconds = 0,
+// From [Matter Application Cluster Specification R1.0], section 1.5.7.4.2
+export const enum DelayedAllOffEffectVariant {
+    /**
+     * Fade to off in 0.8 seconds
+     */
+    Fade = 0,
     NoFade = 1,
-    "50PercentDimDownIn_0p8SecondsThenFadeToOffIn_12Seconds" = 2,
+    /**
+     * 50% dim down in 0.8 seconds then fade to off in 12 seconds
+     */
+    DimDownThenFade = 2,
 }
-const OnOffDelayedAllOffEffectVariantT = { tlvType: TlvType.UnsignedInt } as Template<OnOffDelayedAllOffEffectVariant>;
+const DelayedAllOffEffectVariantT = { tlvType: TlvType.UnsignedInt } as Template<DelayedAllOffEffectVariant>;
 
-export const enum OnOffDyingLightEffectVariant {
-    "20PercentDimUpIn_0p5SecondsThenFadeToOffIn_1Second" = 0,
+// From [Matter Application Cluster Specification R1.0], section 1.5.7.4.2
+export const enum DyingLightEffectVariant {
+    /**
+     * 20% dim up in 0.5s then fade to off in 1 second
+     */
+    DimUpThenFade = 0,
 }
-const OnOffDyingLightEffectVariantT = { tlvType: TlvType.UnsignedInt } as Template<OnOffDyingLightEffectVariant>;
+const DyingLightEffectVariantT = { tlvType: TlvType.UnsignedInt } as Template<DyingLightEffectVariant>;
 
-const OnOffEffectVariantT = UnsignedIntT as Template<OnOffDyingLightEffectVariant | OnOffDelayedAllOffEffectVariant>;
+// From [Matter Application Cluster Specification R1.0], section 1.5.7.4.2
+const EffectVariantT = UnsignedIntT as Template<DyingLightEffectVariant | DelayedAllOffEffectVariant>;
 
 /*
+// TODO Add Bitfield Type
   <bitmap name="OnOffControl" type="BITMAP8">
     <cluster code="0x0006"/>
     <field name="AcceptOnlyWhenOn" mask="0x01"/>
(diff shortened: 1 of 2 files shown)


src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+         * The OnWithTimedOff command allows devices to be turned on for a specific duration with a guarded off
+         * duration so that SHOULD the device be subsequently switched off, further OnWithTimedOff commands, received
+         * during this time, are prevented from turning the devices back on.
+         */
+        onWithTimedOff: Command(0x42, OnWithTimedOffRequestT, 0x42, NoResponseT), /* optional: true */
     },
 );
+
+/*
+Features:
```
Apollon77, line 108: This might more go to the CLuster Featires in Server implementation, also mentioned there

src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+
+import {
+    OnOffCluster, OnOffDelayedAllOffEffectVariant, OnOffDyingLightEffectVariant, OnOffEffectIdentifier
+} from "../OnOffCluster";
+import { ClusterServerHandlers } from "./ClusterServer";
+import { AttributeServer } from "./AttributeServer";
+
+/*
+TODO: Global Cluster fields needs to be added also here because, as discussed, based on the implementation.
```
Apollon77, line 14: Global Cluster Fields - but depend on other clusters if they are available/implemented - also with §circular" deps

src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+
+/*
+TODO: If the Scenes server cluster is implemented on the same endpoint, the following extension field SHALL
+      be added to the Scene Table:
+      * OnOff
+ */
+
+export const OnOffClusterHandler: () => ClusterServerHandlers<typeof OnOffCluster> = () => ({
+    // TODO In fact for the Cluster server implementation we need "initialization code" or thinks like this.
```
Apollon77, line 37: I need somehow an "instance" representation of the handlers ... so in fact would be the ClusterServer instance or such

src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,160 @@
...
+            }
+            if (value === 0) {
+                offWaitTime.set(0);
+                onOff.set(false);
+            }
+            if (value > 0 && onOff.get()) {
+                // TODO This is the 100% accurate impelementation right now ...
+                //      maybe could be optimized without a "100ms timeout loop",
+                //      but for now should do exactly what specs say
+                this.onTimeTimeout = setTimeout(() => {
```
Apollon77, line 58: We need to do it that way if the onTime/offWaitTime values always need to be accurate when "get()" them. Alternative would be to be able to override the get implementation of an Attribute instance to calculate the current "rest time" value on get

src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,160 @@
...
+            }
+            if (value > 0 && !onOff.get()) {
+                this.offWaitTimeout = setTimeout(() => {
+                    this.offWaitTimeout = null;
+                    offWaitTime.set(value - 1);
+                }, 100);
+            }
+        });
+    },
+    // TODO The trick that we used in Device.ts to access "optionalAttributes" ... how to do it here?
```
Apollon77, line 79: I (again) need access here to globalSceneControl, but I was not able to get the trick you showed me with "WithOptionalAttributed" to work here

src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,160 @@
...
+                        break;
+                }
+                break;
+        }
+    },
+    onWithRecallGlobalScene: async ({ attributes: { onOff, globalSceneControl } }) => {
+        if (globalSceneControl.get()) {
+            return; // discard command
+        }
+        // TODO the Scene cluster server on the same endpoint SHALL recall its global scene, updating the OnOff
```
Apollon77, line 143: Needs acces to scene cluster on same Endpoint (as soon as we implement it)

src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,160 @@
...
+            onOff.set(false);
+            onTime.set(0);
+        } else {
+            onOff.set(true);
+            if (onTime.get() === 0) {
+                offWaitTime.set(0);
+            }
+        }
+    },
+    offWithEffect: async ({request: { effectId, effectVariant}, attributes: { onOff, globalSceneControl } }) => {
```
Apollon77, line 103: Needs access to level Cluster (as soon as we implement it)

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+*/
+
+const OffWithEffectRequestT = ObjectT({
+    effectId: Field(0, OnOffEffectIdentifierT),
+    effectVariant: Field(1, OnOffEffectVariantT),
+});
+
+const OnWithTimedOffRequestT = ObjectT({
+    onOffControl: Field(0, UnsignedIntT), /* TODO: Type is a Bit map! */
+    onTime: Field(1, UnsignedIntT), /* min:0, max: 254 */
```
mfucci, line 52: This is nullable

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+    offWaitTime: Field(2, UnsignedIntT), /* min:0, max: 254 */
```
mfucci, line 53: This is nullable but the specs is not clear what null means

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+ * Attributes and commands for switching devices between 'On' and 'Off' states.
+ */
 export const OnOffCluster = Cluster(
     0x06,
     "On/Off",
     {
-        on: Attribute(0, BooleanT),
+        onOff: Attribute(0, BooleanT, false), /* reportable: true, scene:true */
+        globalSceneControl: OptionalAttribute(0x4000, BooleanT, true),
+        onTime: OptionalWritableAttribute(0x4001, UnsignedIntT, 0),
```
mfucci, line 65: This is nullable but the specs is not clear what null means

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+        offWaitTime: OptionalWritableAttribute(0x4002, UnsignedIntT, 0),
```
mfucci, line 66: This is nullable but the specs is not clear what null means

src/matter/cluster/server/OnOffServer.ts
mfucci: I am not sure why you need to add a listener on this attribute:

1. onTime.set(value): outside the OnWithTimedOff state sets the value but does nothing

2. onTime.set(value): during the OnWithTimedOff state modifies the value and this is used to know how long is left to wait before turning off (yes, with a 100ms accuracy, but I think this is fine given for the purpose of this cluster...)

For point 1, maybe there should be a wait to prevent set() to change the value outside a OnWithTimedOff cycle

src/matter/cluster/server/OnOffServer.ts
mfucci: I would recommend to implement the scene control in a 2nd pass...

src/matter/cluster/server/OnOffServer.ts
mfucci: Let's try if passing other clusters in the constructor would work.
At the end, the device (in the Matter definition) will be responsible for instantiating all cluster serves and orchestrating them so it can pass other clusters instances when starting up.

src/matter/cluster/server/OnOffServer.ts
mfucci: Sounds good

mfucci (commented):
I would suggest to split this PR even further:
1. Cluster definition update
2. Skeleton for the server implementation
3. One PR for each command implementation, including an unit test for it (maybe even one integration test?)

We are starting to be close for 1 but still a lot of things to decide on for 2 and 3

Apollon77:
> I would suggest to split this PR even further:

Agreed for #1 and #3. Lets discuss #2 in our call later. Because of also "Initialization code" needed and such and some stuff interacts with each other I do not think that this makes sense. In fact for me it is a "develop the logics together.

But yes, this "dummy PR" is basically exactly for this sync :-)

src/matter/cluster/server/OnOffServer.ts
Apollon77: cool ... exactly this was the "missing puzzle piece"

src/matter/cluster/server/OnOffServer.ts
Apollon77: The reason I need it is because if on "on" command comes in then the offWaitTime is rest to 0 and such cases. So also other commands interact with onTime and offWaitTime.
So my idea was to have the logic in exactly one place - that means that the "attribute value handler" handles it based on the value which is set. So no other code need to care about the details.

The alternative would be to start and stop the Timer in several commands and this would be then spread everywhere (ok, mainly "stopping timer").

src/matter/cluster/server/OnOffServer.ts
Apollon77: For sure - can only be implemented when Scenes are finished, so leaving as Todo for now. Was mainly to tell that we need the inter-cluster-access

src/matter/cluster/server/OnOffServer.ts
Apollon77: In fact also "second implementation step". But already here this way will not work that good nbecause LevelControl cluster needs OnOff Cluster instance and OnOff CLuster needs Level Control cluster because they _both_ interact with each other. So Would only work if we modify config after instanciation :-)
Maybe a topic for the call tonight

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { BooleanT } from "../../codec/TlvObjectCodec";
-import { Attribute, Cluster, Command, NoArgumentsT, NoResponseT } from "./Cluster";
+import { BooleanT, Field, ObjectT, StringT, Template, UnsignedIntT, UnsignedLongT } from "../../codec/TlvObjectCodec";
+import { Attribute, OptionalAttribute, OptionalWritableAttribute, Cluster, Command, NoArgumentsT, NoResponseT } from "./Cluster";
+import { TlvType } from "../../codec/TlvCodec";
 
+export const enum OnOffStartUpOnOff {
```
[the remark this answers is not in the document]
Apollon77: Yes specs 1.5.5.1 (Application)
> If the previous value of the OnOff attribute is equal to FALSE, set the OnOff attribute to TRUE. If the previous value of the OnOff attribute is equal to TRUE, set the OnOff attribute to FALSE (tog­ gle).

:-)

I will add spec refs

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+    Off = 0,
+    On = 1,
+    Toggle = 2,
+}
+const OnOffStartUpOnOffT = { tlvType: TlvType.UnsignedInt } as Template<OnOffStartUpOnOff>;
+
+export const enum OnOffEffectIdentifier {
```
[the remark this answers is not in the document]
Apollon77: Ok, i fact right ... here Chip XML (https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/onoff-cluster.xml#L20) and specs also use other names .. grrmppfff .... Will sync with specs rather then chiptool XML

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+    DelayedAllOff = 0,
+    DyingLight = 1,
+}
+const OnOffEffectIdentifierT = { tlvType: TlvType.UnsignedInt } as Template<OnOffEffectIdentifier>;
+
+export const enum OnOffDelayedAllOffEffectVariant {
+    FadeToOffIn_0p8Seconds = 0,
```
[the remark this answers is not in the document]
Apollon77: was from the chiptool XML ... but yes I can shorten them ... The specs do not provide "names "in these cases

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+/**
+ * Attributes and commands for switching devices between 'On' and 'Off' states.
+ */
 export const OnOffCluster = Cluster(
     0x06,
     "On/Off",
     {
-        on: Attribute(0, BooleanT),
+        onOff: Attribute(0, BooleanT, false), /* reportable: true, scene:true */
+        globalSceneControl: OptionalAttribute(0x4000, BooleanT, true),
```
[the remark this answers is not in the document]
Apollon77: Could be an idea, then also the "global features" would be "static" (on is lighting , one not). Good idea.

Maybe I need then info on how to "code" that with the typers, but second step

Apollon77:
Ok, I will rework the code here and update the PR that we do nor leave reference to the review points. WHen Ok I will withdraw the PR and open split ones for cluster specs and implementation. Rest we talk in our sync later

src/matter/cluster/server/OnOffServer.ts
Apollon77: lets talk in our call ... I miss how I can access the attributes from the cluster in the handler "constructor"

src/matter/cluster/OnOffCluster.ts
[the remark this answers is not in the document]
Apollon77: adjusted and added comment with the details

src/matter/cluster/OnOffCluster.ts
```diff
@@ -4,18 +4,111 @@
...
+
+/**
+ * Attributes and commands for switching devices between 'On' and 'Off' states.
+ */
 export const OnOffCluster = Cluster(
     0x06,
     "On/Off",
     {
-        on: Attribute(0, BooleanT),
+        onOff: Attribute(0, BooleanT, false), /* reportable: true, scene:true */
```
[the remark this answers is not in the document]
Apollon77: meehh TS2474: const enum member initializers can only contain literal values and other computed enum values. :-)

Apollon77:
PS: For the "Nullable" for onTime and offWaitTime there in fact is a definition in the specs in 1.5.7.6.4. of the Application Specs:

> If the values of the OnTime and OffWaitTime attributes are both not equal to null, the server SHALL then update these attributes every 1/10th second until both the OnTime and OffWaitTime attributes are equal to 0, as follows: ...

this case should only happen in rare cases because in many other cases the attributes are set to 0 or to "minimum of two values" (null becomes also 0 then?) ... and in fact 0 will have the same meaning from how I understand the description.

But there is also something I do not understand:
in one place specs say

> The OffWaitTime attribute specifies the length of time (in 1/10ths second) that the ‘Off’ state SHALL be guarded to prevent another OnWithTimedOff command turning the server back to its ‘On’ state (e.g., when leaving a room, the lights are turned off but an occupancy sensor detects the leaving person and attempts to turn the lights back on). 

And later the logic is:

> If the value of the OnOff attribute is equal to TRUE and the value of the OnTime attribute is greater than zero, the server SHALL decrement the value of the OnTime attribute. If the value of the OnTime attribute reaches 0, the server SHALL set the OffWaitTime and OnOff attributes to 0 and FALSE, respectively.

That means how I read it that as soon as the "onTime" is over the light is turned off, But also the offWaitTime is set to 0 ... aka it renders it irreklevant because will not "Block" anything anymore because set to 0 ...

Apollon77:
I close this one ... after the cluster PRs are merged I will open clusterserver PRs to iterate on the implementation details

PR #86, opened 2022-10-26
Add support for optional commands

mfucci:


Apollon77:
Improvise to also add a usage of that into the pr kind of as proof for it to work. 

Maybe add the optional OnOff command specs from my OnOff draft PR but do not have an implementation. Because that should now work, right?!

Else for such new types it is a challenge to know it is working as expected. 

But yes in general (can not really judge the type magic) LGTM

Apollon77 (approved):

mfucci:
That's what I did to test that this is working:
1. added an optional command on OnOffCluster
2. figured out that it was not working
3. dug to nail down to a bug in typescript compiler
4. figured out a workaround (Merge)

I deleted the optional command in OnOffCluster before submitting the PR because this should not go in.
Well, maybe next time I can create a fake branch that you can checkout to see what is the behavior. Since it is mostly compile-time only, the only place where you can test is in the IDE

Apollon77:
I was sure that you tested it ... so all great!

PR #93, opened 2022-10-28
Add EnumT template to simplify dealing with enums

mfucci:


Apollon77:
You saw the TS warmings/errors:

> Conversion of type 'Template<number>' to type 'Template<T>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

Apollon77 (approved):

PR #94, opened 2022-10-28
Fix merge and make mandatory

mfucci:
Merge was using an variable AB that could overridden when used
Make mandatory was erasing the writable type


Apollon77 (approved):
Well ... magic ;-) (and too late today to get this intomy head honestly) ;-)

PR #95, opened 2022-10-28
Add strong type support for VendorId and NodeId

mfucci:
This is kind of a hacky way of doing it: force casting the raw type to a specialized type.

The clean way would have been to introduce a method that the deserializer can use to transform to raw type to the internal typed version.
If you don't like the hacky way, let me know and I can do the clean version.
We will probably need to the clean version at some point...


Apollon77 (approved):
pragmatic fine for me too ... you are the "types expert" :-)

PR #97, opened 2022-10-29
Add constraints bytestream array

mfucci:


Apollon77 (approved):
LGTM. very helpful. I can offer to check the specs TODOs  you added if you like

mfucci:
**Specs TODOs**: we should probably add the reference to the specs everywhere and we can check the constraints at the same time. Probably not a top priority though...

Apollon77:
I will put that on my list when I check all the "root node" stuff ...

PR #98, opened 2022-10-29
Support event definitions

mfucci:
Add support for event declarations in clusters.
There is no implementation associated with them yet.


Apollon77:
Even if not yet asked for review I had aminute between garden works ... I really like that structure!!

mfucci:
Thanks! I think I just forgot to add you as a reviewer but it is ready for review!

Apollon77 (approved):

Apollon77:
Here you go :-)
With all these PRs from today I will adjust all my stuff. 

I need to fight today a bit with Amazon, so might need a day for that

PR #100, opened 2022-10-29
Add jsdocs for BasicInformationCluster

mfucci:
If you like the format for defining the cluster, let's use it as a model for the other cluster definitions.


Apollon77:
Yes mkes sense ... I wil update all next week in that form

Apollon77 (approved):

PR #104, opened 2022-10-31
Add support for bitmap in data models

mfucci:
Usage:
```
BitMapT({
    flag1: Bit(3),
    flag2: Bit(5),
})
```

Probably many features will be missing when actually using it, like how to handled reserved bits that should always be 1, non-boolean values (using multiple bits), etc...

### DIFF
--- a/src/matter/cluster/Cluster.ts
+++ b/src/matter/cluster/Cluster.ts
@@ -41,8 +41,8 @@ export const enum EventPriority {
 }
 export interface Event<T> { id: number, template: Template<T>, priority: EventPriority, optional: boolean }
 export interface OptionalEvent<T> extends Event<T> { optional: true }
-export const Event = <FT extends FieldTemplates>(id: number, priority: EventPriority, data: FieldTemplates = {}): Event<TypeFromFieldTemplates<FT>> => ({ id, template: ObjectT(data), priority, optional: false });
-export const OptionalEvent = <FT extends FieldTemplates>(id: number, priority: EventPriority, data: FieldTemplates = {}): Event<TypeFromFieldTemplates<FT>> => ({ id, template: ObjectT(data), priority, optional: true });
+export const Event = <FT extends FieldTemplates>(id: number, priority: EventPriority, data: FT = <FT>{}): Event<TypeFromFieldTemplates<FT>> => ({ id, template: ObjectT(data), priority, optional: false });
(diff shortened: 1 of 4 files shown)


src/matter/cluster/Cluster.ts
```diff
@@ -41,8 +41,8 @@ export const enum EventPriority {
 }
 export interface Event<T> { id: number, template: Template<T>, priority: EventPriority, optional: boolean }
 export interface OptionalEvent<T> extends Event<T> { optional: true }
-export const Event = <FT extends FieldTemplates>(id: number, priority: EventPriority, data: FieldTemplates = {}): Event<TypeFromFieldTemplates<FT>> => ({ id, template: ObjectT(data), priority, optional: false });
-export const OptionalEvent = <FT extends FieldTemplates>(id: number, priority: EventPriority, data: FieldTemplates = {}): Event<TypeFromFieldTemplates<FT>> => ({ id, template: ObjectT(data), priority, optional: true });
+export const Event = <FT extends FieldTemplates>(id: number, priority: EventPriority, data: FT = <FT>{}): Event<TypeFromFieldTemplates<FT>> => ({ id, template: ObjectT(data), priority, optional: false });
```
mfucci, line 44: That was a but in a previous PR but now ts compiler is complaining about...

src/matter/interaction/InteractionClient.ts
```diff
@@ -131,7 +131,7 @@ export class InteractionClient {
             if (result !== undefined) {
                 const resultCode = result.result.code;
                 if (resultCode !== ResultCode.Success) throw new Error(`Received non-success result: ${resultCode}`);
-                if (responseTemplate !== NoResponseT) throw new Error("A response was expected for this command");
+                if ((responseTemplate as any) !== NoResponseT) throw new Error("A response was expected for this command");
```
mfucci, line 134: Not sure why the ts compiler is now complaining here and what is its problem.
It says that there is no overlap of responseTemplate and NoResponseT but this is not true.

responseTemplate can be Template<void> for Command<?, void>

Well, this type check is not important , so casting to any to avoid the none-sense problem. This won't have any runtime impact.

Apollon77 (approved):
I think for now "normal" bits are enought

mfucci:
I added an unit test so you can see how to use them there,

PR #105, opened 2022-10-31
small updates for BasicInformationCluster and add documents

Apollon77:
As a follow up to your #100 here is a small update for BasicInformationCluster to sync on two topics.
Additionally it adds the other two basic documents to the Specification file (will be used soon)

### DIFF
--- a/src/matter/cluster/BasicInformationCluster.ts
+++ b/src/matter/cluster/BasicInformationCluster.ts
@@ -11,13 +11,13 @@ import { AccessLevel, Attribute, Cluster, Event, EventPriority, OptionalAttribut
 import { MatterCoreSpecificationV1_0 } from "../../Specifications";
 
 
-/** 
+/**
  * Provides constant values related to overall global capabilities of this Node, that are not cluster-specific.
- * 
+ *
  * @see {@link MatterCoreSpecificationV1_0} § 11.1.6.2
  */
 const CapabilityMinimaT = ObjectT({
-    /** Indicates the minimum number of concurrent CASE sessions that are sup­ported per fabric. */
+    /** Indicates the minimum number of concurrent CASE sessions that are supported per fabric. */
     caseSessionsPerFabric: Field(0, BoundedUnsignedIntT({ min: 3 })),
 
     /** Indicate the actual minimum number of concurrent subscriptions supported per fabric. */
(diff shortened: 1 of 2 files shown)


src/matter/cluster/BasicInformationCluster.ts
```diff
@@ -28,29 +28,36 @@ const CapabilityMinimaT = ObjectT({
...
+ *
  * @see {@link MatterCoreSpecificationV1_0} § 11.1
  */
 export const BasicInformationCluster = Cluster({
     id: 0x28,
     name: "Basic Information",
 
+    /* TODO We need to add the global Attributes/Elements, see Core Spec 7.13 - with new structure maybe like that?
+            And yes some attributes could be automatically generated from the registered command implementations or such.
+    globalClusterAttributes: {
```
Apollon77, line 40: What do you think of that? I could already prepare this like that when updating the others ... name is tbd, but I think with the new stricture it would be the "way to go" to do it that way :-)

src/matter/cluster/BasicInformationCluster.ts
```diff
@@ -28,29 +28,36 @@ const CapabilityMinimaT = ObjectT({
...
+        clusterRevision: 1
+    },
+    */
+
     /** @see {@link MatterCoreSpecificationV1_0} § 11.1.6.3 */
     attributes: {
         /** Revision number of the Data Model against which the Node is certified */
-        dataModelRevision: Attribute(0, UnsignedIntT),
+        dataModelRevision: Attribute(0, BoundedUnsignedIntT({ min: 0, max: 0xFFFF })),
```
Apollon77, line 48: Proposal: All uint8 and uint16 types we do that way? uint32 stay as UnsignedIntT and 64bit is already an other type?
I know that this is no topic in encoding, but so we enforce the types as in specs ...

What do you think?

src/matter/cluster/BasicInformationCluster.ts
```diff
@@ -59,15 +66,15 @@ export const BasicInformationCluster = Cluster({
...
 
-        /** Human readable representation of the {@link BasicInformationCluster.attributes.hardwareVersion hardwareVersion} attribute. */
+        /** Human-readable representation of the {@link BasicInformationCluster.attributes.hardwareVersion hardwareVersion} attribute. */
         hardwareVersionString: Attribute(8, StringT({ minLength: 1, maxLength: 64 })),
 
         /** Current version number for the software running on this Node. A larger value is newer than a lower value. */
         softwareVersion: Attribute(9, UnsignedIntT, { default: 0 }),
 
-        /** Human readable representation of the {@link BasicInformationCluster.attributes.softwareVersion softwareVersion} attribute. */
+        /** Human-readable representation of the {@link BasicInformationCluster.attributes.softwareVersion softwareVersion} attribute. */
```
Apollon77, line 77: But I think slowly we shpuld thik about maximum text length! Else these "one line comments" can get very strange".

src/matter/cluster/BasicInformationCluster.ts
mfucci: **global attributes**: I think we need to add "revision" part of the cluster definition, all other global attributes are generated from the data already specified here so don't need to be declared.

Add part of my TODO list of things to do. I would remove from this PR since it won't be done this week and won't require those declarations

src/matter/cluster/BasicInformationCluster.ts
mfucci: I think we need to define UInt8T, UInt16T, UInt32T and UInt64T and remove UnsignedIntT, BoundedUnsignedIntT, UnsignedLongT and BoundedUnsignedLongT.

min: 0 is already done since those values are unsigned
number maximum is higher the UInt32 so need to be bounded, bigint maximum is higher than UInt64 so also needs to be bounded.

I will send a PR in a 30mn to add this.

src/matter/cluster/BasicInformationCluster.ts
mfucci: Let's worry about maximum text length per line when we add lint.
Using one line Jsdoc makes the code quite readable but I agree @link makes the line longer...

src/matter/cluster/BasicInformationCluster.ts
Apollon77: Ok, Then I move it into a COmment above - easier to convert later when the "infos" are already collected

src/matter/cluster/BasicInformationCluster.ts
Apollon77: I saw. I prepare  my side so that I can refresh PR when new types are existing ...

src/matter/cluster/BasicInformationCluster.ts
Apollon77: moved to docs of the cluster in the file so that info is not lost

src/matter/cluster/BasicInformationCluster.ts
Apollon77: (or as proposed .. we leave, merge and I send cleanup PRs, then we remove the old types). Up to you

mfucci (approved):
Fine if you merge now, I should be able to merge those changes in my branch

Apollon77:
Ok,, merged

PR #106, opened 2022-10-31
update Pase and Case messages according to specs

Apollon77:


Apollon77:
@mfucci  Do you have any idea why he errors on length? https://github.com/mfucci/node-matter/pull/106/checks#step:6:18 ... no length defined in what I see

Apollon77:
Locally the error says:

```
2022-9-2 1:56:40.000 INFO MatterController Paired with device: Matter end-to-end device
2022-9-2 1:56:40.000 ERROR InteractionServerMessenger Error: Parameter declaration is longer than 256. Length: 541, value: 0x3082021906092a864886f70d010702a082020a30820206020103310d300b06096086480165030402013082017106092a864886f70d010701a08201620482015e152400012501f1ff3602050080050180050280050380050480050580050680050780050880050980050a80050b80050c80050d80050e80050f80051080051180051280051380051480051580051680051780051880051980051a80051b80051c80051d80051e80051f80052080052180052280052380052480052580052680052780052880052980052a80052b80052c80052d80052e80052f80053080053180053280053380053480053580053680053780053880053980053a80053b80053c80053d80053e80053f80054080054180054280054380054480054580054680054780054880054980054a80054b80054c80054d80054e80054f80055080055180055280055380055480055580055680055780055880055980055a80055b80055c80055d80055e80055f80056080056180056280056380182403162c04135a494732303134325a423333303030332d32342405002406002507942624080018317d307b020103801462fa823359acfaa9963e1cfa140addf504f37160300b0609608648016503040201300a06082a8648ce3d04030204473045022024e5d1f47a7d7b0d206a26ef699b7c9757b72d469089de3192e678c745e7f60c022100f8aa2fa711fcb79b97e397ceda667bae464e2bd3ffdfc3cced7aa8ca5f4c1a7c
      1) the client commissions a new device
      2) the session is resumed if it has been established previously
    subscription
      3) subscription sends updates when the value changested in 3ms
```

Apollon77:
AAHHHHHHHHH ...

```
export const ByteStringT = (constraints: ArrayConstraints = { maxLength: 256 }):Template<Buffer> => ({ tlvType: TlvType.ByteString, validator: arrayValidator(constraints) });

```

So yes ... 256 :-)

I check which PR introduced it ... I think simply needs to be removed or?

Apollon77:
Was introduced in "clean-basic-info-cluster". I try to remove it

Apollon77:
@mfucci tests green again with removed maxLength ... In many places in specs there is "No" limit ... I check what this means now again

Apollon77:
Octent String length in Core Specs 7.18.2.26. says:

It is RECOMMENDED to define constraints on the maximum possible string length.

:-)) Means for me ... nothing defined, no length set, so should be correct like here

src/matter/session/secure/CaseMessages.ts
```diff
@@ -15,56 +23,79 @@ export const KDFSR3_INFO = Buffer.from("Sigma3");
...
         activeRetransTimeoutMs: OptionalField(2, UnsignedIntT),
     })),
 });
 
+/** @see {@link MatterCoreSpecificationV1_0} § 4.13.2.3 */
 export const CaseSigma2ResumeT = ObjectT({
-    resumptionId: Field(1, ByteStringT()),
-    resumeMic: Field(2, ByteStringT()),
-    sessionId: Field(3, UnsignedIntT),
+    resumptionId: Field(1, ByteStringT({ length: 16 })),
```
mfucci, line 70: Maybe those are generic enough to be extract as ByteString16T -> this was reduce the minimal memory footprint since the same object will be used everywhere.

src/matter/session/secure/CaseMessages.ts
```diff
@@ -15,56 +23,79 @@ export const KDFSR3_INFO = Buffer.from("Sigma3");
...
+    resumeMic: Field(2, ByteStringT({ length: 16 })),
+    sessionId: Field(3, BoundedUnsignedIntT({ min: 0, max: 0xFFFF })),
 });
 
+/** @see {@link MatterCoreSpecificationV1_0} § 4.13.2.3 */
 export const CaseSigma3T = ObjectT({
-    encrypted: Field(1, ByteStringT({ maxLength: 400 })), // TODO: check max length in specs
+    encrypted: Field(1, ByteStringT()),
```
mfucci, line 77: This has an implicit maximal length somehow.
Encrypted data are: newOpCert | intermediateCACert | signature
So 400 + 400 + 64 + 16 (auth tag length) = 880

Maybe the best way to handle that is to create an Encrypted() method that will automatically create a ByteStringT with the right max.

src/matter/session/secure/CaseMessages.ts
```diff
@@ -15,56 +23,79 @@ export const KDFSR3_INFO = Buffer.from("Sigma3");
...
 });
 
+/** @see {@link MatterCoreSpecificationV1_0} § 4.13.2.3 */
 export const SignedDataT = ObjectT({
-    newOpCert: Field(1, ByteStringT({ maxLength: 400 })), // TODO: check max length in specs
+    newOpCert: Field(1, ByteStringT()),
```
mfucci, line 82: Name is actually incorrect, should be called nodeOpCert. This is the data passed to the method: addOperationalCert of cluster OperationalCredentials, so the max is 400.

Maybe a nodeOperatrionCertT const can be created and used everywhere?

src/matter/session/secure/CaseMessages.ts
```diff
@@ -15,56 +23,79 @@ export const KDFSR3_INFO = Buffer.from("Sigma3");
...
 });
 
+/** @see {@link MatterCoreSpecificationV1_0} § 4.13.2.3 */
 export const EncryptedDataSigma2T = ObjectT({
-    newOpCert: Field(1, ByteStringT({ maxLength: 400 })), // TODO: check max length in specs
+    newOpCert: Field(1, ByteStringT()),
     intermediateCACert: OptionalField(2, ByteStringT()),
-    signature: Field(3, ByteStringT()),
-    resumptionId: Field(4, ByteStringT()),
+    signature: Field(3, ByteStringT({length: CRYPTO_GROUP_SIZE_BYTES * 2})),
```
mfucci, line 92: Maybe use a signatureLength = 64 (or CRYPTO_GROUP_SIZE_BYTES * 2)

src/matter/session/secure/CaseMessages.ts
```diff
@@ -15,56 +23,79 @@ export const KDFSR3_INFO = Buffer.from("Sigma3");
 export const TBE_DATA2_NONCE = Buffer.from("NCASE_Sigma2N");
 export const TBE_DATA3_NONCE = Buffer.from("NCASE_Sigma3N");
 
+/** @see {@link MatterCoreSpecificationV1_0} § 3.5.1 */
```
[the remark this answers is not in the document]
Apollon77: I will move tomorrow morning (my time) - I have also no issue if you adjust it in the PR ...

src/codec/TlvObjectCodec.ts
```diff
@@ -66,7 +66,7 @@ const arrayValidator = ({ minLength = 0, maxLength, length }: ArrayConstraints)
 export const Constraint = <T,>(template: Template<T>, validator: (value: T, name: string) => void): Template<T> => ({...template, validator });
 export const StringT = (constraints: ArrayConstraints = { maxLength: 256 }):Template<string> => ({ tlvType: TlvType.String, validator: arrayValidator(constraints) });
 export const BooleanT:Template<boolean> = { tlvType: TlvType.Boolean };
-export const ByteStringT = (constraints: ArrayConstraints = { maxLength: 256 }):Template<Buffer> => ({ tlvType: TlvType.ByteString, validator: arrayValidator(constraints) });
+export const ByteStringT = (constraints: ArrayConstraints = {}):Template<Buffer> => ({ tlvType: TlvType.ByteString, validator: arrayValidator(constraints) });
```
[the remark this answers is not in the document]
Apollon77: Hm ... The specs define limits and when they do we utilize them, so all "without limits" in the spacs actually should be expected as "without a limit" maybe it could be an idea to use 2k or something like this as a first higher default ... then it will be just a matter of time when we need to increase stuff again.

I understand the concerns, but in fact - defining a limit when specs do not define one is critical in my eyes because it can only go wrong.

On the other had we are still Javascript, so all memory is allocated dynamically as needed and these limits here are "just for type/content checking" ... no idea if limiting it here "unneededly" is really bringing any benefit

src/matter/session/secure/CaseMessages.ts
```diff
@@ -15,56 +23,79 @@ export const KDFSR3_INFO = Buffer.from("Sigma3");
...
+
+/** @see {@link MatterCoreSpecificationV1_0} § 4.13.2.3 */
 export const CaseSigma1T = ObjectT({
-    random: Field(1, ByteStringT()),
-    sessionId: Field(2, UnsignedIntT),
-    destinationId: Field(3, ByteStringT()),
-    ecdhPublicKey: Field(4, ByteStringT()),
-    mrpParams: OptionalField(5, ObjectT({
+    random: Field(1, ByteStringT({ length: 32 })),
+    sessionId: Field(2, BoundedUnsignedIntT({ min: 0, max: 0xFFFF })),
```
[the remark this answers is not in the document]
Apollon77: Ok, will try it tomorrow

src/matter/session/secure/CaseMessages.ts
```diff
@@ -15,56 +23,79 @@ export const KDFSR3_INFO = Buffer.from("Sigma3");
...
+    destinationId: Field(3, ByteStringT({ length: CRYPTO_HASH_LEN_BYTES })),
+    ecdhPublicKey: Field(4, ByteStringT({ length: CRYPTO_PUBLIC_KEY_SIZE_BYTES })),
+    mrpParams: OptionalField(5, ObjectT({ // sed-parameter-struct
```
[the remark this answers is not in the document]
Apollon77: Ok I can do that ... was not before so the comment was just a left over from my to check all types and structs.

src/matter/session/secure/CaseMessages.ts
Apollon77: Ok I can scan over all code tomorow and see ... If then 12 and 32 would be both

src/matter/session/secure/CaseMessages.ts
Apollon77: If you say that this is the limit ... I checked docs and were not able to explicitely find these number in the Case specs ... but maybe also just because I'm not that deep in there.

Will put the 880 as limit tomorrow

Extra type I would leave up to you ;-)

src/matter/session/secure/CaseMessages.ts
Apollon77: will rename

src/codec/TlvObjectCodec.ts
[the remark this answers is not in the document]
mfucci: My point was: I don't think no limit ever makes sense for a protocol supposed to low computing devices.

Should I refactor this to force to specify a maxLenght on each ByteString so there is more default?
The C++ code a max on each bytestring...

src/matter/session/secure/CaseMessages.ts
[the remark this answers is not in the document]
mfucci: I will write a PR now to unblock this.

src/matter/session/secure/CaseMessages.ts
mfucci: I will try to do a PR now to add this method.

src/codec/TlvObjectCodec.ts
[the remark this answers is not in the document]
Apollon77: Which limit does C++ use? (And I hope they verify in all these places that the incoming or generated content is never longer then these "unspecified" maximum lengths, else they have funny buffer overflows and security issues ...Ok This should not be with us on JavaScript side ...

Effectively ... I'm ok with readding that "defaukt maxLength" with lets say 1024 bytes and we will see when the first one run against it and where :-)
As said effectively we just use this as verificator and it has nothing to do with memotry allocation, or? Or is the maxLength used anywhere as Buffer size (which btw would also be wrong because it is just the maximum allowed size). So if not then the Buffer will allocate memory "as needed" content wise ... So no idea if t is really comparable to the C++ side

src/matter/session/secure/CaseMessages.ts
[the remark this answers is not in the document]
Apollon77: @mfucci it is even "better" ... this struct is used by both Pase and Case ... SO ideally we should put it somewhere else. Described it is in 2.12.5 which is "Data model" scope. Any idea where to best put it?

src/codec/TlvObjectCodec.ts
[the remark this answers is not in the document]
Apollon77: I adjusted the PR with 1024 and at least tests run

src/matter/session/secure/CaseMessages.ts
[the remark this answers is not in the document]
Apollon77: As proposed in other PR: Leave old types and "derecate" them, lets merge stuff and I will do a run over codebase for that and provide PR. So it would be not blocking each other

src/matter/session/secure/CaseMessages.ts
Apollon77: * length 32 -> 9x in code base
* length 16 -> 5x in code base
* then some more with "constants" that end in 64 or 65 ...

Not sure if it is worth own types :-) What do you think?

But would also see this as a "post PR" topic

Apollon77:
ok

Apollon77:
@mfucci I also updated this PR to be working with master,so can also be merged

src/matter/session/secure/CaseMessages.ts
Apollon77: maybe lets handle that really in follow up PR because for now it is proofed to work that way

src/matter/session/secure/PaseMessages.ts
```diff
@@ -6,40 +6,53 @@
...
+});
+
+/** @see {@link MatterCoreSpecificationV1_0} § 4.13.1.2 */
 export const PbkdfParamRequestT = ObjectT({
-    random: Field(1, ByteStringT()),
-    sessionId: Field(2, UInt16T),
-    passcodeId: Field(3, UInt16T),
+    random: Field(1, ByteStringT({ length: 32 })),
+    sessionId: Field(2, UInt16T), // Specs: range: 16bits
+    passcodeId: Field(3, UInt16T), // Specs: length: 16bits so min is 0x1000 ??
```
Apollon77, line 27: Any idea on that? ;-)

mfucci: I don't think we have to worry about that. When used, 1 will be converted to 0x0001.

mfucci (approved):

src/matter/session/secure/PaseMessages.ts
Apollon77: Ok, lets see how it will end ... my interpretation of "length 16 bit" was that at least 16 needs to be used .. this in theory brings me to min 0x8000 (uups typo in comment) aka 32.768 as "smallest number that is actually using 16 bit in length

PR #109, opened 2022-11-01
Add definitions and partial refactoring

mfucci:
Sorry not finished yet... I need to go to bed...


Apollon77:
Alternative: leave the old types in and I do the cleanup in follow up pr and then we remove old types when done. Will not need long.

Apollon77:
Would the "Bound" also work to e.g. limit a GroupId (aka 16 bit number 0..max) in some cases to 1..max? Because we need this in some places. If yes that would be awesome.

As said in other PRs: Make it easy and just use the new types in 1-2 places ... and leave old types in and I fix all of that (because if have that done all locally already - no need that you work through specs again - I did yesterday whole day kind of)

mfucci:
Finished now! I had to finish the updating all the integer types to have it compiling.
I found some bugs as well that I fixed.

Apollon77:
wow you are up early :-) Checking now ... will check then when merging my local stuff in (yeeayyy - that will be fun) ;-))

Apollon77 (dismissed):
LGTM ... I will recheck all when merging with my local changes again

mfucci:
I fixed UInt64 maximum. I took Int64 max instead of UInt64 and it crashed when I tried against iOs 16.1

Apollon77 (approved):

Apollon77:
approved

mfucci:
Sorry, after merging I need your Appoval stamp again before I can merge :-)

Apollon77 (approved):
sure, and here you go

PR #118, opened 2022-11-07
Add AccessControlCluster specs

Apollon77:


src/matter/cluster/AccessControlCluster.ts
```diff
@@ -0,0 +1,177 @@
...
+    },
+
+    /** @see {@link MatterCoreSpecificationV1_0} § 9.10.7 */
+    events: {
+        /**
+         * The cluster SHALL send AccessControlEntryChanged events whenever its ACL attribute data is changed by an
+         * Administrator.
+         * @see {@link MatterCoreSpecificationV1_0} § 9.10.7.1
+         */
+        accessControlEntryChanged: Event(0, EventPriority.Info, {
```
mfucci, line 149: To avoid repeating the common part with the next event, this can be generalized:
```
const AccessChangeEvent = <T>(entryTemplate: Template<T>) => ({
            /** The Node ID of the Administrator that made the change, if the change occurred via a CASE session. */
            adminNodeID: Field(0, NodeIdT), /* nullable: true */
            /** The Passcode ID of the Administrator that made the change, if the change occurred via a PASE session. */
            adminPasscodeID: Field(1, Bound(UInt16T,{ min: 0, max: 0xFFFF })), /* nullable: true */
            /** The type of change as appropriate. */
            changeType: Field(2, EnumT<ChangeTypeEnum>()),
            /** The latest value of the changed entry. */
            latestValue: Field(3, entryTemplate), /* nullable: true */
});
```

mfucci (approved):

PR #120, opened 2022-11-07
Add BridgedDeviceBasicInformationCluster specs

Apollon77:
This one is a bit special because it is a case of a "derived" cluster.

In fact it is derived from basicInformationCluster, BUT The optional/mandatory/"should exist at all" conformance and some other stuff is different from the parent definition.
Thats why for now I did a copy and defined anything as expected ...

How we should handle those?

### DIFF
--- a/src/matter/cluster/BridgedDeviceBasicInformationCluster.ts
+++ b/src/matter/cluster/BridgedDeviceBasicInformationCluster.ts
@@ -1,0 +1,92 @@
+/**
+ * @license
+ * Copyright 2022 The node-matter Authors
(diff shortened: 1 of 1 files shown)


mfucci:
Fine to duplicate.

An alternative is to modify the attributes of basicInformationCluster to make them optional.
I did the opposite in ClusterServer.ts to make attributes mandatory so if you want to have fun with typing, you can write MakeAttributesOptional...

mfucci (approved):

Apollon77:
In fact it is more then just optional/not optional becasue in fact also "should not exist", so leave duplicated for now

PR #123, opened 2022-11-07
Add GroupsCluster specs

Apollon77:


src/matter/cluster/GroupsCluster.ts
```diff
@@ -0,0 +1,178 @@
...
+    EnumT, BitMapT, Bit
+} from "../../codec/TlvObjectCodec";
+import { Attribute, Cluster, Command, NoArgumentsT, NoResponseT } from "./Cluster";
+import { StatusCode } from "../interaction/InteractionMessages";
+import { GroupIdT } from "../common/GroupId";
+import { MatterApplicationClusterSpecificationV1_0 } from "../../Specifications";
+
+/** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.3.7.1 */
+const AddGroupRequestT = ObjectT({
+    groupId: Field(0, GroupIdT), /* min: 1 */
```
mfucci, line 22: Should "min: 1" be part of GroupId ?
Or has GroupId = 0 a special meaning?

src/matter/cluster/GroupsCluster.ts
```diff
@@ -0,0 +1,178 @@
...
+/** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.3.7.3 */
+const GetGroupMembershipRequestT = ObjectT({
+    groupList: Field(0, ArrayT(GroupIdT)), /* groupId min: 1 */
+});
+
+/** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.3.7.9 */
+const GetGroupMembershipResponseT = ObjectT({
+    /** contain the remaining capacity of the Group Table of the node. */
+    capacity: Field(0, UInt8T), /* nullable: true */
+    groupList: Field(1, ArrayT(GroupIdT)),
```
mfucci, line 53: Can GroupId be 0 here?

mfucci (approved):

src/matter/cluster/GroupsCluster.ts
Apollon77: you are right, fixing

src/matter/cluster/GroupsCluster.ts
Apollon77: bit I add min as Bound now

mfucci: Ok, let's keep it like that and fix it later when we rework Templates

mfucci (approved):

PR #124, opened 2022-11-07
Add ScenesCluster specs

Apollon77:


src/matter/cluster/ScenesCluster.ts
```diff
@@ -0,0 +1,295 @@
...
+import { Attribute, OptionalAttribute, Cluster, Command, OptionalCommand, NoResponseT } from "./Cluster";
+import { StatusCode } from "../interaction/InteractionMessages";
+import { GroupIdT } from "../common/GroupId";
+import { ClusterIdT } from "../common/ClusterId";
+import { NodeIdT } from "../common/NodeId";
+import { AttributeIdT } from "../common/AttributeId";
+import { MatterApplicationClusterSpecificationV1_0 } from "../../Specifications";
+
+/* TODO
+<bitmap name="ScenesCopyMode" type="BITMAP8">
```
mfucci, line 29: Bitmaps are implemented now:
const ScenesCopyMode = BitmapT({
   copyAllScenes: Bit(0),
});

src/matter/cluster/ScenesCluster.ts
```diff
@@ -0,0 +1,295 @@
...
+         */
+        nameSupport: Attribute(4, nameSupportBitmapT, { default: { sceneNames: true } }),
+        /** Holds the Node ID (the IEEE address in case of Zigbee) of the node that last configured the Scene Table. */
+        lastConfiguredBy: OptionalAttribute(5, NodeIdT), /* nullable: true */
+    },
+
+    /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.4.9 */
+    commands: {
+        /**
+         * @see {@link MatterApplicationClusterSpecificationV1_0} § 1.4.9.2
```
mfucci, line 239: nit: move spec ref after the jsdoc
Same below

mfucci (approved):

src/matter/cluster/ScenesCluster.ts
Apollon77: missed this one. added now

src/matter/cluster/ScenesCluster.ts
```diff
@@ -0,0 +1,293 @@
...
+import { GroupIdT } from "../common/GroupId";
+import { ClusterIdT } from "../common/ClusterId";
+import { NodeIdT } from "../common/NodeId";
+import { AttributeIdT } from "../common/AttributeId";
+import { MatterApplicationClusterSpecificationV1_0 } from "../../Specifications";
+
+/** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.4.6.1 */
+const AttributeValuePairT = ObjectT({
+    /**
+     * This field SHALL be present or not present, for all instances in the Scenes cluster. If this field is
```
mfucci, line 31: :-) This field SHALL be present or not present : difficult not to respect this rule, I think all instances in the Scenes cluster will respect it.

src/matter/cluster/ScenesCluster.ts
```diff
@@ -0,0 +1,293 @@
...
+     * not present, then the data type of AttributeValue SHALL be determined by the order and data type defined
+     * in the cluster specification. Otherwise the data type of AttributeValue SHALL be the data type of the
+     * attribute indicated by AttributeID.
+     */
+    attributeId: OptionalField(0, AttributeIdT),
+
+    /** This is the attribute value as part of an extension field set. */
+    attributeValue: Field(1, ArrayT(UInt32T)), /* TODO: type wise would be "ArrayT(unknown)", but ... :-) */
```
mfucci, line 39: If the type is not known but will be known at runtime, you should use AnyT

This expect the caller to already have transform this field using the right template before.
See DataReportT in InteractionMessage.ts as an example

src/matter/cluster/ScenesCluster.ts
Apollon77: I removed the "optional" for now ... because if fieldname is optional the the whole entry makes no sense.

src/matter/cluster/ScenesCluster.ts
Apollon77: adjusted

src/matter/cluster/ScenesCluster.ts
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * @see {@link MatterApplicationClusterSpecificationV1_0} § 1.4.6.2
+ */
+export const ExtensionFieldSetT = ObjectT({
+    clusterId: Field(0, ClusterIdT),
+    attributeValueList: Field(1, ArrayT(AttributeValuePairT)),
+});
+
+/** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.4.9.2 */
+/** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.4.9.9 */
```
mfucci: I think there can only be one JsDoc per element to be valid.

```suggestion
/** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.4.9.2 and 1.4.9.9 */
```

mfucci (dismissed):

Apollon77:
Thank you!

PR #125, opened 2022-11-07
Onoff Cluster rework

Apollon77:
As discussed ... updated OnOffCLuster with rename including moving the bare minimum implementation into a server

### DIFF
--- a/src/matter/cluster/server/OnOffServer.ts
+++ b/src/matter/cluster/server/OnOffServer.ts
@@ -25,10 +25,10 @@ export const OnOffClusterHandler: () => ClusterServerHandlers<typeof OnOffCluste
     on: async ({ attributes: { onOff } }) => {
         onOff.set(true);
     },
-    off: async ({ attributes: { onOff} }) => {
+    off: async ({ attributes: { onOff } }) => {
         onOff.set(false);
     },
-    toggle: async ({ attributes: { onOff} }) => {
+    toggle: async ({ attributes: { onOff } }) => {
         if (onOff.get()) {
             onOff.set(false);
         } else {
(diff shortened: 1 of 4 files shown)


src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,37 @@
...
+* GeneratedCommandList: empty
+* EventList: empty
+* FabricIndex: empty
+ */
+
+export const OnOffClusterHandler: () => ClusterServerHandlers<typeof OnOffCluster> = () => ({
+    on: async ({ attributes: { onOff } }) => {
+        onOff.set(true);
+    },
+    off: async ({ attributes: { onOff} }) => {
```
mfucci, line 28: nit: space after onOff.
Same below

src/matter/cluster/server/OnOffServer.ts
```diff
@@ -0,0 +1,37 @@
...
+* AttributeList:
+  * onOff
+* AcceptedCommandList:
+  * Supported always: Off, On, Toggle
+* GeneratedCommandList: empty
+* EventList: empty
+* FabricIndex: empty
+ */
+
+export const OnOffClusterHandler: () => ClusterServerHandlers<typeof OnOffCluster> = () => ({
```
mfucci, line 24: This doesn't need to be a function since there is no configuration:
export const OnOffClusterHandler: ClusterServerHandlers<typeof OnOffCluster> = {

mfucci (dismissed):

src/matter/cluster/server/OnOffServer.ts
Apollon77: Yes I know, but it will be as soon as we add the next PR when we implement more logic (okk then it even need to be  a class/instance as discussed, thats why I left it that way. I can also change it. as you like

mfucci (approved):

PR #131, opened 2022-11-20
Add index.ts to expose public library classes

mfucci:


Apollon77 (commented):
Yes this is a minimal start but I think noone could build e.g. a "Device" with this. When I look at https://github.com/mfucci/node-matter/blob/main/src/Device.ts what we all need to build "our minimum test device" then you see which other classes ou all need that are not yet exported officially ... Or do I miss something?

mfucci:
I think importing classes directly from sub-directories should work as well:
import { OnOffCluster } from "node-matter/dist/matter/cluster/OnOffCluster";

The goal of this PR is to start exposing something and have people playing with it...

Apollon77:
I thought that ... but why not expose the directories as exports? like net, cluster and so on? But if it should be bare minimum then this PR make a first small step  ;-)

Apollon77 (approved):

mfucci:
Exporting each directory is not really recommended: we did this for matter.js initially but the result was not great so we moved towards directly exposing every class from the index.ts.

PR #137, opened 2022-12-05
Groups implementation (partly)

Apollon77:
Hey @mfucci,

as promised here a simple case ... Groups implementation ... I need a instance to store the groups assignments (and later also needs to be persisted).

Additionally it would also makle sense to be able to call other impelemtations on the same instance to avod duplicate code - or I need a way to add additional methds to that instance to split out common used code.

Lets start with this :-)


Apollon77:
BTW: Yes I also tried the idea of 

```
class GroupsClusterHandler implements ClusterServerHandlers<typeof GroupsCluster> {
}
```

That you provided but then I was lost on how to access the request parameters (and in next step for other clusters also the cluster attributes)

mfucci:
Updated your branch with an implementation suggestion using a closure to create an internal state (this is the old fashion js class design pattern...)

I tried to create a real class using:
class GroupsClusterHandler implements ClusterServerHandlers<typeof GroupsCluster> {
   [...]
}

that does work but the code looks really ugly since typescript compiler requires to declare args type on each methods, and since it is auto-generated from the ClusterDefinition, it is ugly.
Using the object notation, ts compiler infers all the correct types, so it is much easier to write / read it.

Apollon77:
Ok, yes so it could work for now ... lets see when we need "initialization "logics"" ;-) ... but you removed the "this->addGroup" call in the addGroupWhenIdentifying method ... what we do there? Or do we duplicate code?

mfucci:
addGroupWhenIdentifying: with return this->addGroup(..), the return signature of addGroupWhenIdentifying was not matching what is described in the cluster defintion.
I let you debug this one...

Let me know when everything is ready for review

Apollon77:
Will check today

Apollon77:
hhmm ... I think there is a spec flaw ... so we need to assume things:

AddGroupIfIdentifying do not have any defined response command and also do not refer to any other. SO my best "assumption" is to use the addresponse response here too. That would contain a status field that we need here and also have one field more as groupId which should be ignored by the receivers if not expected.

Adjusted PR again.

Apollon77:
BTW for the Group case /unicast and such topic in the Todos here I digged a bit ... 

Core Specs 4.14 described Group communication.
> Multicast addressing is accomplished using the 16-bit Group ID field as the destination address. A multicast group is a collection of Nodes, all registered under the same multicast Group ID.

In fact it is a special type of encrypted session where the communication runs over

Core Specs 4.15 described the Group Key management logic for the relevant encryption

src/matter/cluster/server/GroupsServer.ts
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+/*
+TODO: Is a "groupcast" the groupId range 0xFF00 - 0xFFFF ??
+ */
+
+export const GroupsClusterHandler: () => ClusterServerHandlers<typeof GroupsCluster> = () => {
+    const clusterGroups = new Map<number, string>();
+
+    const addGroupLogic = (groupId: GroupId, groupName: string): TlvAddGroupResponse => {
```
Apollon77, line 40: I think that way we can avoid code duplication when following that schema ... I just had no success to make the return type vorrect here (but hopefully easy one for you) :-)

Apollon77:
@mfucci Maybe we are closer now here?

mfucci:
> hhmm ... I think there is a spec flaw ... so we need to assume things:
> 
> AddGroupIfIdentifying do not have any defined response command and also do not refer to any other. SO my best "assumption" is to use the addresponse response here too. That would contain a status field that we need here and also have one field more as groupId which should be ignored by the receivers if not expected.
> 
> Adjusted PR again.

Turon told me that the CHIP codebase is the source of truth, even above the specs. So when something is not clear, just look at  what is done there...

Apollon77:
lol, good to know ... will try if I get something out there

Apollon77:
@mfucci @turon Any starting point? I tried to find it in the chip codebase but failed ...

mfucci:
> @mfucci @turon Any starting point? I tried to find it in the chip codebase but failed ...

Here?
https://github.com/project-chip/connectedhomeip/blob/af8161cb092ec01e184338cd0598fcd63d4481c2/src/app/clusters/groups-server/groups-server.cpp#L340

Apollon77:
Ok with this (thank you for the link ... I did not discovered that) my assumption was right. Response is like addgroup with a status field.

Apollon77:
Oooohhhh nooooo that's not right. I'm fact they have a default response defined which is only returning a status. 

emberAfSendImmediateDefaultResponse(status) ... that means that the NoResponse type should get an optional status?! Or we need an extra "defaultResponseWithStatus" type

Apollon77:
Ok, updated here too ... so only topic left is the "is it a "groupcast or not" topic which I already found out in https://github.com/mfucci/node-matter/pull/137#issuecomment-1346098733, but that is a more "core" topic because it starts with data from this incoming udp.

@mfucci  What do you think, merge with the Todos or leave open? (if second I would maybe bring in the cluster specs part in a separate PR)

mfucci (dismissed):

src/matter/cluster/server/ClusterServer.ts
```diff
@@ -20,7 +21,7 @@ export type AttributeInitialValues<A extends Attributes> = Merge<Omit<{ [P in Ma
 type MandatoryCommandNames<C extends Commands> = {[K in keyof C]: C[K] extends OptionalCommand<any, any> ? never : K}[keyof C];
 type OptionalCommandNames<C extends Commands> = {[K in keyof C]: C[K] extends OptionalCommand<any, any> ? K : never}[keyof C];
 type AttributeGetters<A extends Attributes> = { [P in keyof A as `get${Capitalize<string & P>}`]?: (session?: Session<MatterDevice>) => AttributeJsType<A[P]> };
-type CommandHandler<C extends Command<any, any>, A extends AttributeServers<any>> = C extends Command<infer RequestT, infer ResponseT> ? (args: { request: RequestT, attributes: A, session: Session<MatterDevice> }) => Promise<ResponseT> : never;
+type CommandHandler<C extends Command<any, any>, A extends AttributeServers<any>> = C extends Command<infer RequestT, infer ResponseT> ? (args: { request: RequestT, attributes: A, session: Session<MatterDevice>, message: Message }) => Promise<ResponseT> : never;
```
Apollon77, line 24: we added it in the other places to pass the message but missed it here in the typing.

Apollon77:
COntent will be in a future matter.js PR

PR #139, opened 2022-12-09
Add QR code generation

mfucci:
This was much harder to write than I thought!

### DIFF
--- a/src/codec/PairingCode.ts
+++ b/src/codec/PairingCode.ts
@@ -1,0 +1,88 @@
+/**
+ * @license
+ * Copyright 2022 The node-matter Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Schema } from "@project-chip/matter.js";
+import { Base38 } from "./Base38";
+import { BitField, BitFieldEnum, BitFlag, ByteArrayBitmapSchema, TypeFromBitmapSchema, BitmapSchema } from "./BitmapSchema";
+import { Verhoeff } from "../math/Verhoeff";
+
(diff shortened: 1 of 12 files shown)


Apollon77:
Wow, will review tomorrow: BTW The chip tool have that also as a link to a Webpage where the "qr code" is in the URL as textual representation ... We can maybe also add such a "URL" to the output later

mfucci:
Done!
Now, when node-matter starts:

1. it prints the QR code
2. Gives the URL to the online version
3. Displays the manual pairing code

src/codec/PairingCode.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    /** Device supports BLE for discovery when not commissioned. */
+    ble: BitFlag(1),
+
+    /** Device is already on the IP network. */
+    onIpNetwork: BitFlag(2),
+});
+
+/** See {@link MatterCoreSpecificationV1_0} § 5.1.3.1 Table 35 */
+const QrCodeDataSchema = ByteArrayBitmapSchema({
```
Apollon77, line 37: Which mechanism adds the 4 bytes at the end of the struct? meaning the "Padding" from specs

Apollon77 (approved):
awesome

lboue:
Can you merge with the master branch if it works fine?

src/codec/PairingCode.ts
mfucci: The ByteArrayBitmapSchema sets to 0 all bits not defined.
Yes, I know: probably we should add a mechanism to explicitly declared reserve bits and their values...

PR #145, opened 2022-12-24
Upgrade matter.js to 0.1.0

mfucci:
This should fix Bug #11
Unknown TLV fields are now ignored in matter,js


Apollon77 (approved):

Apollon77:
fingers crossed :-) And great christmas time

PR #149, opened 2022-12-27
Add IPv6 support

mfucci:


src/matter/common/MessageExchange.ts
```diff
@@ -182,17 +182,21 @@ export class MessageExchange<ContextT> {
...
     }
 
-    private async retransmitMessage(message: Message, retransmissionCount: number) {
-        await this.channel.send(message);
-        retransmissionCount++;
-        if (retransmissionCount === this.retransmissionRetries) return;
-        this.retransmissionTimer = Time.getTimer(this.activeRetransmissionTimeoutMs, () => this.retransmitMessage(message, retransmissionCount))
-            .start();
+    private retransmitMessage(message: Message, retransmissionCount: number) {
+        this.channel.send(message)
```
Apollon77, line 186: why you start using such then/-catch constructs and not try/catch with await?

src/util/Ip.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * @license
+ * Copyright 2022 The node-matter Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ByteArray } from "@project-chip/matter.js";
+
+export function isIPv4(ip: string) {
+    return ip.includes(".")
```
Apollon77, line 10: nice pragmatic solution preventing a regex :-)

Apollon77 (approved):
Wow, great progress and work. From my perspective it makes sense ... just a comment for try/catch vs then/catch ... up to you

src/matter/common/MessageExchange.ts
mfucci: When an error is throw in an async method, it crashes node.js with an error message: an error was thrown in an async method and not caught.
It doesn't tell you what was the error or where it was thrown.

I spent one hour adding logs until I found where was the error.

To avoid this: all "roots" of async methods should handle async errors.
retransmitMessage was previously an async method, but called from a Timer, which doesn't handle async errors. Instead I removed async from this method and added the handling for the async errors in it.

Ideally, the compiler should raise an error when an async method is called in a synchronous context, but I don't see any options for this.

src/util/Ip.ts
mfucci: Quick and dirty :-)
This doesn't ensure that the string is a well-formed IPv4 address but good enough when the IP comes from another API.

src/matter/common/MessageExchange.ts
Apollon77: Yes I know that ... but why you use send.then(...).catch(...) and not

```
try {
  send();
  ...
} catch(e) {
  ...
}
```
The affect is the same ... in fact it is "blowing up the code a bit" (because try/catch is more lines) vs "strange promise chaining (that is the same as the callback hell that promises wanted to solve) ;-)

But just my opinion.

Yes both work ... just question about code style

mfucci: You are right, it is the same at the end.
My preference would still be to ensure that an async entry point is not async itself...
Maybe there is a better to achieve this.

Apollon77: >  My preference would still be to ensure that an async entry point is not async itself...

What you mean with this?

mfucci: I meant that all async entry points are not async themselves.

If there a method:
async doSomethingAsync() {}

It should either be called inside an async method or by an async entry point, which is not async.

An example of an entry poitn could be:
start() {
  doSomethingAsync()
      .catch(error => handleError(error));
}

I should probably create a AsyncRunner that takes care of the error handling and add fix the stacktrace so the non-async portion is added after the async portion which will make it much easier to debug...

Apollon77: In fact yes all right ... The call of an async method in an not async endpoint requires exactly what you did, but leaves us with "no proper error handling at all" (unless we start having Event handlers everywhewre and so "forward" such error as events. Additionally caklling an async method in a non async one means it is really executed async (so "parallel" to other code ... this can also produce a fair amount of issues (because it was not "awaited" you can not be sure (ok, it is 100% sure that it is __not__) the logic is really executed already when the method returns.

So when all this is ok (code is executed async after method returned AND no real error handling) then that approach is fine ...

PR #157, opened 2022-12-31
Update ID / name to match the certificate

mfucci:


Apollon77 (approved):
Hm ... so we also need to generate the certificate based on settings like name and vendor ids and such?

mfucci:
Yes, ideally all the certificates should be auto-generated...
But at least IDs should match with what is declared.

PR #168, opened 2023-01-06
Fix incorrect parameter maximum

mfucci:
The specs says 2047 in page 728 but this is incorrect.
The discriminator is an unsigned 12-bit so the max is 4095 and iOS sends discriminators above 2047.


Apollon77 (approved):

Apollon77:
fixled it as docs bug https://github.com/project-chip/connectedhomeip/issues/24301

PR #171, opened 2023-01-06
Temporary hack to subscribe only to OnOff

mfucci:
Ugly hack to only subscribe to OnOff attribute when the caller subscribe to */*/*
This allows iOs to operate node-matter, but unfortunately that doesn't do the trick for Android.


Apollon77 (approved):
Comment should state todo. ;-) But fine.

Apollon77:
In fact the reality would be to subscribe to anything

PR #180, opened 2023-01-07
Allow Cluster specs to return status where relevant

Apollon77:


mfucci:
Can you give me more clues on how this is used and the relevant part of the specs describing it?

This change will completely change the return signature of the impacted commands... unlikely to be correct.

More details:
A normal command that returns a response will generate a TlvInvokeResponse containing:
response: {
    path: {
       endpointId,
       clusterId,
       id,
    },
    response: [TLV-encoded response],
},

A command that has a TlvNoResponse declared, will return nothing. In this case TlvInvokeResponse will contain a result instead a response (different field in the TLV structure):
result: {
     path: {
       endpointId,
       clusterId,
       id,
    },
    result: {
       code: ResultCode enum
   }
}

TlvStatusOnlyResponse will still generate a response even if status is undefined. The response will be an empty object, so TlvInvokeResponse will contain a response instead of a result.

Did you mean to change the value of the ResultCode instead? Right now it is hardcoded as Success but you can create a specifial Error with other ResultCodes if needed to fill the result.code with an error code.

Maybe ResultCode should be the same enum as StatusCode. I wrote this part of the code before the specs were available...

Apollon77:
Thats basically a very good question :-)

Bildschirm­foto 2023-01-09 um 07 35 44

I refered to this part of the Core Specs and thoight it will be a normal response ... 

Coes that shed the needed light for you? If it should be the "response code" then tell menhow to set an errr and I will adjust

Apollon77:
Additionally I can find this in the specs ...

Bildschirm­foto 2023-01-09 um 07 43 55

mfucci:
I think what is implemented now is "Y" for TlvNoResponse: status is returned
Do you see any commands with "N" in the currently implemented specs?

InteractionMessages.ts#126 => probably the fields here should be rename with the clear matching location in the specs:

```
/** See @see {@link MatterCoreSpecificationV1_0} § 8.8.3.1 */
export const TlvInvokeResponse = TlvObject({
    suppressResponse: TlvField(0,  TlvBoolean),
   
    /** See @see {@link MatterCoreSpecificationV1_0} § 10.5.13 */
    invokeResponses: TlvField(1, TlvArray(TlvObject({
         /** See @see {@link MatterCoreSpecificationV1_0} § 10.5.12 */
        command: TlvOptionalField(0, TlvObject({
            commandPath: TlvField(0, TlvList({
                endpointId: TlvField(0, TlvUInt16),
                clusterId: TlvField(1, TlvUInt32),
                id: TlvField(2, TlvUInt32),
            })),
            commandFields: TlvField(1, TlvAny),
        })),

         /** See @see {@link MatterCoreSpecificationV1_0} § 10.5.14 */
        status: TlvOptionalField(1, TlvObject({
            commandPath: TlvField(0, TlvList({
                endpointId: TlvField(0, TlvUInt16),
                clusterId: TlvField(1, TlvUInt32),
                id: TlvField(2, TlvUInt32),
            })),
            status: TlvField(1, TlvObject({
                status: TlvField(0, TlvUInt16),
                clusterStatus: TlvOptionalField(1, TlvUInt16),
            })),
        })),
    }))),
    interactionModelRevision: TlvField(0xFF, TlvUInt8),
});
```

Apollon77:
> I think what is implemented now is "Y" for TlvNoResponse: status is returned

Ok, but an hard coded status? Or how do I return a status with "TlvNoResponse" or "TlvNoResponse needs to get an optional status to return?
I tried to return { status: ... } but this gaves a type error because he expected no return at all 

>  Do you see any commands with "N" in the currently implemented specs?
I can have a look again (would do that when going through this PR to fix it according how needed) but out of my memories only cases with "N" were the ones where an other response was defined ... AND cases in group cluster and such where it is described that "no response should be sent if request was sent as group-cast" (but thats topic to tackle later I assume)

Apollon77:
Ok this PR is complete bullshit as we learned ... So we still need a way to return a "different status then "OK"" for TlvNoResponse ...

PR #182, opened 2023-01-09
Ignore incorrect ack

mfucci:
Ignore incorrect ack
Change code flow to have more accurate retransmission delays

### DIFF
--- a/src/matter/common/MessageExchange.ts
+++ b/src/matter/common/MessageExchange.ts
@@ -14,6 +14,7 @@ import { Time, Timer } from "../../time/Time";
 import { Logger } from "../../log/Logger";
 import { NodeId } from "./NodeId";
 import { ByteArray } from "@project-chip/matter.js";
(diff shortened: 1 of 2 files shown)


Apollon77 (approved):
I'm not that 100% in that promise resolving code, but let's go ahead. Me and home assistant had that issue, so we will see it

mfucci:
Same here, not 100% sure it will fix the issue since it is not easy to reproduce...
But I tried this code and it is not breaking the current flows...

JimBuzbee:
I did a bit of testing this morning. Previously when I had run node-matter on a raspberry-pi in the far reaches of my house with spotty wifi, I could easily get the incorrect ack crash. But this morning, with the new build, I have not seen it, even when hitting the On/Off .Google Home button as fast as I can. 

Nice work! I'll keep an eye on it today.

mfucci:
Thanks! Closing the issue then!

Apollon77:
I also verified it and yes I did not got that errror again yesterday

PR #187, opened 2023-01-11
Ad getter handler and currentFabricIndex

mfucci:
Unfortunately that doesn't solve the pairing issue with iOS. I will dig more tomorrow.

### DIFF
--- a/src/matter/fabric/FabricManager.ts
+++ b/src/matter/fabric/FabricManager.ts
@@ -18,6 +18,10 @@ export class FabricManager {
         return new FabricIndex(this.fabrics.length);
     }
 
+    removeFabric(fabricIndex: FabricIndex) {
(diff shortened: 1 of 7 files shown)


src/matter/fabric/FabricManager.ts
```diff
@@ -18,6 +18,10 @@ export class FabricManager {
         return new FabricIndex(this.fabrics.length);
     }
 
+    removeFabric(fabricIndex: FabricIndex) {
+        this.fabrics.splice(this.fabrics.findIndex(fabric => fabric.fabricIndex.index === fabricIndex.index), 1);
```
Apollon77, line 22: Here you assume that the the array idex is the same as fabric Index, right? can we do this?

Maybe better to re-set fabrict with a "filtered" array where yoiu just remove the entry with the relevant index?

Apollon77 (approved):
ignore last comment. I clearly need another coffee ... approved

PR #192, opened 2023-01-16
Allow Hierarchical Endpoint structures

Apollon77:
In preparateion to implement Agrregator and Bridged Devcies the Endpoiint structure needs to be more hierarchical to easiely allow structures like shown in Core Specs pg 476 (and following).

All data, and especially the partsList is "bubbled" fropmbottom to top as requitred by the specs.

The next step after the PR ws merged is a second PR that adjusts InteractionServer to use it instead of the current flat logic.

The tests show already how it would "look like" later then basically and cover the "current" simply endpoint case as well as some more complex ones taken from the Bridged Nodes examples.

In this case I also found two "bugs" (OptionalWriteableAttribute - already made also PR to matter.js, and naming for the BridgedDevice cluster).

I also verified that with this code (assuing InteractionServer changes from next step done alteady too) still works for the current simple endpoint usecase on Alexa

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /build
 /node_modules
+/.idea
--- a/src/matter/cluster/BridgedDeviceBasicInformationCluster.ts
+++ b/src/matter/cluster/BridgedDeviceBasicInformationCluster.ts
@@ -20,7 +20,7 @@ import { MatterCoreSpecificationV1_0, TlvBoolean, TlvField, TlvString, TlvString
  *
  * @see {@link MatterCoreSpecificationV1_0} § 9.13
  */
-export const BasicInformationCluster = Cluster({
+export const BridgedDeviceBasicInformationCluster = Cluster({
     id: 0x39,
     name: "Bridged Device Basic Information",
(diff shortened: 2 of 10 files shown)


Apollon77:
PS: The PR also introduce that the "devicesList" in the descriptior needs to be able to be an array because e.g. when adding Bridged Devices an endpoint needs to have multiple device types

Apollon77:
@mfucci PR updated to latest main branch with your fabric changes

Apollon77:
@mfucci PR updated baed on your comments/questions and questions also answered

the DEVICE -> DEVICES I would do in my followup PR because else gets too messy locally :-)

Apollon77:
PS: after https://github.com/mfucci/node-matter/pull/193 bridge pairing with my local version based on this PR  is also working on ios

Apollon77:
@mfucci Please maybe also have alook at https://github.com/Apollon77/node-matter/pull/1 ... this is then the "followup" to use the new structure ... it got a bit strange with now branch on branch because I still work on my fork :-(

PR #193, opened 2023-01-17
Revert incorrect change from admincomm-cluster-rework 

Apollon77:
which caused ios pairing issues. By reverting this pairing on ios work again - devices stil unresponsive for me, but pairing works

### DIFF
--- a/src/matter/cluster/AdminCommissioningCluster.ts
+++ b/src/matter/cluster/AdminCommissioningCluster.ts
@@ -70,10 +70,6 @@ const TlvOpenBasicCommissioningWindowRequest = TlvObject({
     commissioningTimeout: TlvField(0, TlvUInt16),
 });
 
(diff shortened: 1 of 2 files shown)


Apollon77:
@mfucci verified, this restores iOS functionality ...

vves:
I just validated that this update allows for the Matter Simulated Device to be added to the iOS Home App on iOS 16.3.  Home App still displays "No Response", but at least commissioning seems to be working again.

Apollon77:
For no response: does it load when you tap the device in home app and make it "big"
In the App? 

Do you have an HomePod?

mfucci (approved):

mfucci:
For Matter devices operations, Apple requires a Apple-compatible hub to be present on the local network, since operating the device is actually done through the cloud and not on the local network (the same for Google, a Google-compatible hub is require for operations).

Issue #189, opened 2023-01-13
Unable to comision device with Smartthings


JimBuzbee:
Failed to add node-matter "device" with Smarthings app. The commision process starts with expected messages after I scan the QR code, then I get a warning about a non-certified device, then the app appears to partially crash sending me back a couple of screens to the scan QR screen again and no device showing up in Smarthings.

[smartthings.txt](https://github.com/mfucci/node-matter/files/10413631/smartthings.txt)

JimBuzbee:
Update - I guess I'm not that familiar with SmartThings because I did find that the device had been added, but it is non-functional. See screenshots attached. 
![Screenshot_20230113-101023](https://user-images.githubusercontent.com/4139136/212378809-03e39941-160b-42aa-abdd-66f9a6a1a125.png)

![Screenshot_20230113-101002](https://user-images.githubusercontent.com/4139136/212378765-1e035a8d-938d-4f37-8172-9a37e2a353c6.png)

JimBuzbee:
One more update - I am getting some interaction between node-matter and Smartthings. I can turn my light on (for me, Govee), but not off.  Every button press in the App sends on "On". Maybe the message "_The device hasn't updated all of its status..._" means that Smarthings doesn't know if the light is on or off so it always sends an "on" ?  Does the spec call for a periodic on/off status message that we are still lacking causing Smarthings to be confused?

Apollon77:
is the above log really complete?

JimBuzbee:
@Apollon77 - It may not be, I probably confuse myself with too many node-matter instances running plus Alexa, Google Home and SmartThings on my net at the same time :-) Attached is new log of what I _think_  is a complete, clean run of adding a new Matter device and then hitting the On/Off button twice in the SmartThings app. You will see two sets of output from "curl" which is what I use to control my Govee lights. Both look like they were told by SmartThings to turn the lights on rather than one "on" and one "off". 


[smartthings1.txt](https://github.com/mfucci/node-matter/files/10415621/smartthings1.txt)

mfucci:
Thanks for reporting this!
There are a few hacky code paths that I am cleaning up right now and I will update on this thread when there is some chance that it might now be fully working.

Apollon77:
I think the relevant part here is

> 2023-0-5 14:37:6.332 ERROR InteractionMessenger Error: Unsupported message type 6

@mfucci I will try to implement it

Apollon77:
[...]

PR #194, opened 2023-01-19
Adds first support for WriteRequests

Apollon77:
should address #189

The support is still pragmatic and several cases are not et checked as by specs, but shoul at least allow to see whats next for "Smartthings pairing".

It also adds more logging in general to get better semantic-readable logs

### DIFF
--- a/src/matter/interaction/InteractionMessenger.ts
+++ b/src/matter/interaction/InteractionMessenger.ts
@@ -94,10 +94,10 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
 
     async handleRequest(
         handleReadRequest: (request: ReadRequest) => DataReport,
+        handleWriteRequest: (request: WriteRequest) => WriteResponse,
         handleSubscribeRequest: (request: SubscribeRequest) => SubscribeResponse | undefined,
         handleInvokeRequest: (request: InvokeRequest) => Promise<InvokeResponse>,
         handleTimedRequest: (request: TimedRequest) => Promise<void>,
-        handleWriteRequest: (request: WriteRequest) => WriteResponse | undefined,
     ) {
         let continueExchange = true;
         try {
@@ -108,6 +108,11 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                     case MessageType.ReadRequest:
                         const readRequest = TlvReadRequest.decode(message.payload);
                         await this.sendDataReport(handleReadRequest(readRequest));
+                        break;
+                    case MessageType.WriteRequest:
+                        const writeRequest = TlvWriteRequest.decode(message.payload);
+                        const writeResponse = handleWriteRequest(writeRequest);
+                        await this.exchange.send(MessageType.WriteResponse, TlvWriteResponse.encode(writeResponse));
                         break;
                     case MessageType.SubscribeRequest:
                         const subscribeRequest = TlvSubscribeRequest.decode(message.payload);
@@ -128,15 +133,6 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                         await handleTimedRequest(timedRequest);
                         await this.sendStatus(StatusCode.Success);
                         continueExchange = true;
-                        break;
-                    case MessageType.WriteRequest:
-                        const writeRequest = TlvWriteRequest.decode(message.payload);
-                        const writeResponse = await handleWriteRequest(writeRequest);
-                        if (writeResponse) {
-                            await this.exchange.send(MessageType.WriteResponse, TlvWriteResponse.encode(writeResponse));
-                        } else {
-                            await this.sendStatus(StatusCode.Success); // ???
-                        }
                         break;
                     default:
                         throw new Error(`Unsupported message type ${message.payloadHeader.messageType}`);
--- a/src/matter/interaction/InteractionMessenger.ts
+++ b/src/matter/interaction/InteractionMessenger.ts
@@ -8,7 +8,20 @@ import { Logger } from "../../log/Logger";
 import { MessageExchange } from "../common/MessageExchange";
 import { MatterController } from "../MatterController";
 import { MatterDevice } from "../MatterDevice";
-import { TlvInvokeRequest, TlvInvokeResponse, TlvReadRequest, TlvDataReport, TlvSubscribeRequest, TlvSubscribeResponse, StatusCode, TlvStatusResponse, TlvTimedRequest, TlvAttributeReport } from "./InteractionMessages";
+import {
+    TlvInvokeRequest,
+    TlvInvokeResponse,
+    TlvReadRequest,
+    TlvDataReport,
+    TlvSubscribeRequest,
+    TlvSubscribeResponse,
+    StatusCode,
+    TlvStatusResponse,
+    TlvTimedRequest,
+    TlvAttributeReport,
+    TlvWriteRequest,
+    TlvWriteResponse
+} from "./InteractionMessages";
 import { ByteArray, TlvSchema, TypeFromSchema } from "@project-chip/matter.js";
 
 export const enum MessageType {
@@ -31,6 +44,8 @@ export type SubscribeResponse = TypeFromSchema<typeof TlvSubscribeResponse>;
 export type InvokeRequest = TypeFromSchema<typeof TlvInvokeRequest>;
 export type InvokeResponse = TypeFromSchema<typeof TlvInvokeResponse>;
 export type TimedRequest = TypeFromSchema<typeof TlvTimedRequest>;
+export type WriteRequest = TypeFromSchema<typeof TlvWriteRequest>;
+export type WriteResponse = TypeFromSchema<typeof TlvWriteResponse>;
 
 const MAX_SPDU_LENGTH = 1024;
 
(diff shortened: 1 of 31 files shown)


src/matter/interaction/InteractionServer.ts
```diff
@@ -252,21 +323,33 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
         return `${endpointName}/${clusterName}/${attributeName}`;
     }
 
+    private getAttribute(path: Partial<Path> ): AttributeWithPath[] {
```
mfucci, line 326: getAttribute and getAttributes methods are confusing because both returns a list of attributes so it unclear was each one does.

I think the intent is this one:
* one will ignore unknown attributes (for read)
* the other one will report an error for unknown attributes (for write)

Maybe it would be better to keep a single method and a an extra argument to indicate which behavior is expected?
undefined can be return for unknown attribute so it can be converted into an error instead of ignoring it.
It is unclear what should the behavior for "star" IDs (undefined) which defines a filter. If no attributes are defined for this filter, should we return an error for the filter or ignore it? Is it even allow to write to a "star" pattern? (Like */34/2)

src/matter/mdns/MdnsBroadcaster.ts
```diff
@@ -30,6 +30,8 @@ export class MdnsBroadcaster implements Broadcaster {
     ) {}
 
     setCommissionMode(mode: number,deviceName: string, deviceType: number, vendorId: VendorId, productId: number, discriminator: number) {
+        logger.debug(`announce commissioning mode ${mode} ${deviceName} ${deviceType} ${vendorId} ${productId} ${discriminator}`);
```
mfucci, line 33: nit: Announce

This method actual doesn't announce but just set the mode. The announce() method should be called to announce...

test/matter/interaction/InteractionProtocolTest.ts
```diff
@@ -51,6 +56,42 @@ const READ_RESPONSE: DataReport = {
...
+        },
+    ],
+    moreChunkedMessages: false,
+};
+
+const WRITE_RESPONSE: WriteResponse = {
+    interactionModelRevision: 1,
+    writeResponses: [
+        {
+            path: {
```
mfucci, line 82: nit: match the declaration above so it is clear which path this is:
path: { endpointId: 0, clusterId: 0x28, attributeId: 100 },

test/matter/interaction/InteractionProtocolTest.ts
```diff
@@ -51,6 +56,42 @@ const READ_RESPONSE: DataReport = {
...
+            data: TlvString.encodeTlv("test"),
+            dataVersion: 0,
+        },
+    ],
+    moreChunkedMessages: false,
+};
+
+const WRITE_RESPONSE: WriteResponse = {
+    interactionModelRevision: 1,
+    writeResponses: [
```
mfucci, line 80: Shouldn't we have a success status for { endpointId: 0, clusterId: 0x28, attributeId: 5 } ?

src/matter/interaction/InteractionServer.ts
```diff
@@ -168,6 +171,74 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
...
+                    status: {
+                        status: StatusCode.UnsupportedWrite
+                    }, // TODO: Find correct status code
+                    path: request.path,
+                });
+            } else {
+                attributes.forEach(({ path, attribute }) => {
+                    // Todo respect ACL and TimedWrite flags
+                    // TODO add checks or dataVersion
+                    if (path.clusterId === BasicInformationCluster.id && path.id !== BasicInformationCluster.attributes.nodeLabel.id) {
```
mfucci, line 209: Can you add more details why this specific attribute should be ignored?
Maybe better to split into 2 MRs: one that adds write support, the other that adds the hack for SmartThings

mfucci (dismissed):
If this code works, I don't want to block moving forward so you can either:
* address my comments
* submit and I will update the code afterwards

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -113,6 +129,15 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                         await this.sendStatus(StatusCode.Success);
                         continueExchange = true;
                         break;
+                    case MessageType.WriteRequest:
+                        const writeRequest = TlvWriteRequest.decode(message.payload);
+                        const writeResponse = await handleWriteRequest(writeRequest);
+                        if (writeResponse) {
+                            await this.exchange.send(MessageType.WriteResponse, TlvWriteResponse.encode(writeResponse));
+                        } else {
+                            await this.sendStatus(StatusCode.Success); // ???
```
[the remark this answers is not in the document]
Apollon77: What about the suppressResponse =true case (which anbot yet implemented but will come). Then there is no response.

test/matter/interaction/InteractionProtocolTest.ts
Apollon77: How I understand the specs we only return write errors. I will add spec reference later.

src/matter/interaction/InteractionServer.ts
Apollon77: Will split it

Apollon77:
I will update the PR and you can review again. I'm fact for smartthings it seems that it do not work correct irrelevant of that fix or not. ;-)

src/matter/interaction/InteractionServer.ts
```diff
@@ -168,6 +171,74 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
         };
     }
 
+    handleWriteRequest(exchange: MessageExchange<MatterDevice>, {suppressResponse, writeRequests }: WriteRequest): WriteResponse | undefined {
+        logger.debug(`Received write request from ${exchange.channel.getName()} for ${writeRequests.length} attributes, suppressResponse=${suppressResponse}`);
+
+        // TODO consider TimedRequest constrains
+
+        const writeResponses = [];
+        for (const request of writeRequests) {
```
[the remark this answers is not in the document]
Apollon77: Do you have an info how to get info if an attribute is writable or the min/maxlength?

src/matter/interaction/InteractionServer.ts
Apollon77: But for context: smartthings sends a write request on basic info Cluster with * as attribute and a text value. According to specs we need to walk through all matching attributes and check the "constrains" and if "writable". Then in fact only the nodeLabel attribute is left to set that value on.

Because of the fact that I could not determine if an attribute is writable and it's length constrains I added that hack to only consider nodeLabel attribute when such a request comes in.

src/matter/interaction/InteractionServer.ts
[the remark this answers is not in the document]
mfucci: set() calls the validator() method internally so you don't have to about min / max or other data constraints

Maybe let's do the same for writable:
* add a writable flag the AttributeServer which is populated at creation with the value from the attribute definition
* check / throw error in the set() method is called on a non-writable attribute (but not on setLocal since the value can be changed by the local code)

If we want to return the correct error Status, we can add an enum in the errors thrown indicating the error reason and convert it into the correct error Status in the catch in the code above.

src/matter/interaction/InteractionServer.ts
mfucci: Ok, let's implement checking if attributes are writable then.
I suggest above to add writable flag in the AttributeServer so we can create a getWritableAttributes method here that will returns only writable attributes.

src/matter/interaction/InteractionMessenger.ts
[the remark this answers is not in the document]
Apollon77: Ok, removed for now and enhcanced TODO in other place

src/matter/interaction/InteractionServer.ts
```diff
@@ -144,18 +145,20 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
...
+    handleReadRequest(exchange: MessageExchange<MatterDevice>, {attributes: attributePaths, isFabricFiltered}: ReadRequest): DataReport {
+        logger.debug(`Received read request from ${exchange.channel.getName()}: ${attributePaths.map(path => this.resolveAttributeName(path)).join(", ")}, isFabricFiltered=${isFabricFiltered}`);
 
         const values = this.getAttributes(attributePaths)
             .map(({ path, attribute }) => {
                 const { value, version } = attribute.getWithVersion(exchange.session);
                 return { path, value, version, schema: attribute.schema };
             });
 
+        logger.debug(`Read request from ${exchange.channel.getName()} resolved to: ${values.map(({ path, value, version }) => `${this.resolveAttributeName(path)}=${Logger.toJSON(value)} (${version})`).join(", ")}`);
```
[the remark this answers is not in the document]
Apollon77: The "previous result is based on the "incoming" attribute list and not on the "real" one after enhancing all the "*"s

src/matter/interaction/InteractionServer.ts
Apollon77: Good ... and how we handle the "constraints"? Reason: Fpr the Writecall that Smartthings is doing also "location" (2 char string) would be writable ... but should be ignored because contrains do not mauth to a "longer then 2 chars node label string" :-)

test/matter/interaction/InteractionProtocolTest.ts
Apollon77: see other discussions

src/matter/interaction/InteractionServer.ts
Apollon77: ignore , I adjust it back :-)

src/matter/interaction/InteractionServer.ts
Apollon77: Ok, I found the issue  I think .... will provide a PR to bring in validator logic

Apollon77: https://github.com/mfucci/node-matter/pull/197

Apollon77: and #198

Apollon77:
PR blocked by PRs #197 and #198 to introduce needed features

Apollon77:
replaced by #215

PR #195, opened 2023-01-19
Request Fabrics dynamically on OperationalCredentialsCluster

Apollon77:
The fabric is removed, but currently the fabric attribute of the cluster is not updated, sonext request still returns the just deleted fabric and confuses controllers.

It is still missing to "kill the secure session" - or at least the "fabric bound on that session", buit I do not know how, so this is still a good first step.

### DIFF
--- a/src/matter/cluster/server/OperationalCredentialsServer.ts
+++ b/src/matter/cluster/server/OperationalCredentialsServer.ts
@@ -52,7 +52,7 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
         }
     },
 
-    addOperationalCert: async ({ request: {operationalCert, intermediateCaCert, identityProtectionKey, caseAdminNode, adminVendorId}, session, attributes: { fabrics } }) => {
+    addOperationalCert: async ({ request: {operationalCert, intermediateCaCert, identityProtectionKey, caseAdminNode, adminVendorId}, session }) => {
         if (!session.isSecure()) throw new Error("addOperationalCert should be called on a secure session.");
         const device = session.getContext();
         const fabricBuilder = device.getFabricBuilder();
@@ -65,7 +65,15 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
         const fabric = await fabricBuilder.build();
         const fabricIndex = device.addFabric(fabric);
 
-        fabrics.setLocal(device.getFabrics().map(fabric => ({
+        // TODO: create ACL with caseAdminNode
+        console.log("addOperationalCert success")
+
+        return { status: OperationalCertStatus.Success, fabricIndex };
+    },
+
+    getFabrics: session => {
+        if (session === undefined || !session.isSecure()) return []; // ???
+        return session.getContext().getFabrics().map(fabric => ({
             fabricId: fabric.fabricId,
             label: fabric.label,
             nodeId: fabric.nodeId,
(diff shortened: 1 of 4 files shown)


src/matter/cluster/server/OperationalCredentialsServer.ts
```diff
@@ -107,14 +107,30 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
...
     },
 
-    removeFabric: async ({ request: {fabricIndex}, session }) => {
+    removeFabric: async ({ request: {fabricIndex}, attributes: {fabrics}, session }) => {
         const device = session.getContext();
-        device.removeFabric(fabricIndex);
-
-        return {status: OperationalCertStatus.Success};
+        if (device.removeFabric(fabricIndex)) {
+            fabrics.setLocal(device.getFabrics().map(fabric => ({
```
mfucci, line 118: Now that this is supported, I might probably be better to make "fabrics" attribute a getterAttribute (value generated on the fly, like "currentFabric", I just added support), rather than an AttributeServer holding a value.
This would avoid to duplicate setting this attribute everywhere to match the current list of fabrics.

src/matter/fabric/FabricManager.ts
```diff
@@ -19,7 +19,10 @@ export class FabricManager {
     }
 
     removeFabric(fabricIndex: FabricIndex) {
-        this.fabrics.splice(this.fabrics.findIndex(fabric => fabric.fabricIndex.index === fabricIndex.index), 1);
+        const index = this.fabrics.findIndex(fabric => fabric.fabricIndex.index === fabricIndex.index);
+        if (!index) return false;
```
mfucci, line 23: Instead of return false, it would be better to throw an Error:
* this makes it obvious what is expected code path versus unexpected path
* there is no not obvious boolean return value from the remove method
* this is consistent with the rest of the codebase: Errors are thrown for unexpected parameters / value sand it is up to the caller to handle them, with a catch-all net in the Messenger to send back a basic error

mfucci (changes requested):

Apollon77:
Will look into both tomorrow

src/matter/fabric/FabricManager.ts
Apollon77: ok

src/matter/cluster/server/OperationalCredentialsServer.ts
Apollon77: Done. pairing with iOS still works

mfucci (approved):

Apollon77:
@mfucci as for the error type in fact two options I see:

1.) we use a RangeError for our case because the id is not found so "out of allowed range". This would reuse an existing error type ;-)

2.) we define a FabricNotExistError extends Error and throw this and can use normal instanceOf for detection. This approach could end up in many special custom internal error types. 

3.) we define a NodeMatterError extends error and add find of a details field to constructor and such. And we define a ErrorDetail enum with allowed sub error types. Then we use throw NodeMatterError(ErrorDetails.FabricNotExist, message) and can catch with instanceOf and check the details info ...
Then we could encapsulate all kids of errors into one object. 
Comparable to https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/error/ZWaveError.ts

What you like/think?

Apollon77:
Beause of abovediscussion I will merge here and we fix in own PR

PR #204, opened 2023-01-26
Separate Path types for attributes and commands

Apollon77:
While woring on WriteRequestsand also other topics I had issues because the commandPath and AttributePaths are using the same txpe but in fact are not "the same structure" in details ... Thats why I splitted these up


mfucci (dismissed):

Apollon77:
@mfucci please re-approve (was the "space" too much)

mfucci (approved):

mfucci (approved):

PR #206, opened 2023-01-26
Forward the Message to invoke command handler

Apollon77:
For Group clusters (and also others) we need to be able to access information from the package header when processing commands. In fact mainly the "request type" (multicast/unicast) and "is it a group request" is relevant based on the current knowledge but it should also not hurt to make all data available.

### DIFF
--- a/src/matter/cluster/server/CommandServer.ts
+++ b/src/matter/cluster/server/CommandServer.ts
@@ -7,6 +7,7 @@
 import { MatterDevice } from "../../MatterDevice";
 import { Session } from "../../session/Session";
 import { TlvSchema, TlvStream } from "@project-chip/matter.js";
+import { PacketHeader } from "../../../codec/MessageCodec";
 
 export const enum ResultCode {
     Success = 0x00,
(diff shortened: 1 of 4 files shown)


src/matter/cluster/server/CommandServer.ts
```diff
@@ -19,12 +20,12 @@ export class CommandServer<RequestT, ResponseT> {
...
         protected readonly responseSchema: TlvSchema<ResponseT>,
-        protected readonly handler: (request: RequestT, session: Session<MatterDevice>) => Promise<ResponseT> | ResponseT,
+        protected readonly handler: (request: RequestT, session: Session<MatterDevice>, packetHeader: PacketHeader) => Promise<ResponseT> | ResponseT,
     ) {}
 
-    async invoke(session: Session<MatterDevice>, args: TlvStream): Promise<{ code: ResultCode, responseId: number, response: TlvStream }> {
+    async invoke(session: Session<MatterDevice>, args: TlvStream, packetHeader: PacketHeader): Promise<{ code: ResultCode, responseId: number, response: TlvStream }> {
         const request = this.requestSchema.decodeTlv(args);
-        const response = await this.handler(request, session);
+        const response = await this.handler(request, session, packetHeader);
```
mfucci, line 28: Let's pass the whole message

mfucci (changes requested):

Apollon77:
@mfucci adjusted as discussed

mfucci (approved):

PR #208, opened 2023-01-28
Introduce MatterJsError class

Apollon77:
This PR introduces an own Error Class (@mfucci after our discussion today I directly called it MatterJsError ... can change that back if too early). it allows to define/transport Error codes/types and also a informational context where the eror happened (yes this is also in tre stack, but maybe helpful also in error message only contexts).

It also adds a first usabe in the removeFabric case as discussed today


test/error/ErrorTest.ts
```diff
@@ -0,0 +1,42 @@
...
+
+describe("Errors", () => {
+
+    context("Test Errors", () => {
+        it("error with code", () => {
+            try {
+                throw new MatterJsError("test", MatterJsErrorCodes.FabricNotFound);
+            } catch (error) {
+                assertMatterJsError(error);
+                assert.equal(isMatterJsError(error), true);
```
AlCalzone, line 24: Unnecessary

test/error/ErrorTest.ts
```diff
@@ -0,0 +1,42 @@
...
+                assert.equal(error.message, "(fabric-not-found): test");
+            }
+        });
+
+        it("error with code and context", () => {
+            try {
+                throw new MatterJsError("test", MatterJsErrorCodes.FabricNotFound, "test-context");
+            } catch (error) {
+                assertMatterJsError(error);
+                assert.equal(isMatterJsError(error), true);
```
AlCalzone, line 35: Dito

test/error/ErrorTest.ts
Apollon77: kk, will remove but yes already dfone by the other assert

mfucci (changes requested):

Apollon77:
@mfucci Ok, I adjusted it to as we discussed today.

We have now:
* MatterError as base
* One example for a specific Error for FabricNotFound
* A new StatusResponseError class with the option go et a specific errors for "all" status Responses that are defined. I tried around a lot today but in the end to have "real" error classes on top level the only way was to define them static. All other options do not allow easy checks like "instanceof FailureStatusResponseError", and I think because the StatusCode list is static it is ok to have it that way
* A tryCatch method as you proposed with the addition that also a "error handler function" can be used to compute the fallback Value, with this we can also react to different kinds of errors and reacht differently

Apollon77:
I also added an Async/Promise variant of the tryCatch method ...

src/error/MatterError.ts
```diff
@@ -0,0 +1,146 @@
...
+import { StatusCode } from "../matter/interaction/InteractionMessages";
+
+/** Error base class for all errors thrown by this library. */
+export class MatterError extends Error {}
+
+/** Specific Error for when a fabric is not found. */
+export class FabricNotFoundError extends MatterError {}
+
+
+/** Error base Class for all errors related to the status response messages. */
```
mfucci, line 16: This is specific to InteractionProtocol so should probably be in matter/interaction
Maybe part of the Messenger?

src/error/MatterError.ts
```diff
@@ -0,0 +1,146 @@
...
+export class StatusResponseError extends MatterError {
+    private static errorMap = new Map<StatusCode, new(...args: any[]) => StatusResponseError>();
```
mfucci, line 18: This is an non-standard way of doing a mapping.

I am not sure where exactly you need the mapping (expect in the test where you can directly grab what you want), but if you do, a more standard way woudl be:
const StatusError: { [statusCode: StatusCode]: StatusResponseError } = {
     [StatusCode.Failure]: FailureStatusResponseError;
     (etc...)
};

Or you can simply add the status code inside the StatusResponseError and not worry about individual naming.

src/error/MatterError.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export class FailsafeRequiredStatusResponseError extends StatusResponseError {}
+StatusResponseError.registerErrorClass(StatusCode.FailsafeRequired, FailsafeRequiredStatusResponseError);
+
+
+export function isMatterError(e: unknown): e is MatterError {
+    return e instanceof MatterError;
+}
+
+type MatterErrorHandler<T> = (error: MatterError) => T | undefined;
```
mfucci, line 112: This should not return undefined

src/error/MatterError.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export function tryCatch<T>(codeBlock: () => T, errorType: typeof MatterError, fallbackValue: MatterErrorHandler<T> | T): T {
```
mfucci, line 114: typeof Error -> there is nothing specifc to a MatterError here

mfucci, line 114: This should proably be moved into a util/ErrorHandle.ts and made generic

mfucci, line 114: nit: fallbackValue -> fallback since this can be a function or a value ?

src/error/MatterError.ts
```diff
@@ -0,0 +1,146 @@
...
+    try {
+        return codeBlock();
+    }
```
mfucci, line 117: nit: catch is usually on the same line
} catch (error) {

src/error/MatterError.ts
```diff
@@ -0,0 +1,146 @@
...
+    catch (error) {
+        if (isMatterError(error) && error instanceof errorType) {
```
mfucci, line 119: I don't think it matter is this is a MatterError or not, should work for any type of error and instandof will already filter on MatterError if errorType extends MatterError

src/error/MatterError.ts
```diff
@@ -0,0 +1,146 @@
...
+    try {
+        return await codeBlock();
+    }
+    catch (error) {
+        if (isMatterError(error) && error instanceof errorType) {
+            if (typeof fallbackValue === "function") {
+                const computedFallbackValue = (fallbackValue as MatterErrorHandler<T>)(error);
+                if (computedFallbackValue !== undefined) return Promise.resolve(computedFallbackValue);
+            } else {
+                return Promise.resolve(fallbackValue);
```
mfucci, line 141: I don't sync you need a Promise.resolve here since the function is declared as async

src/matter/cluster/server/OperationalCredentialsServer.ts
```diff
@@ -109,16 +110,18 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
     removeFabric: async ({ request: {fabricIndex}, session }) => {
         const device = session.getContext();
 
-        try {
+        const status = tryCatch(() => {
```
mfucci, line 113: Nit: indentation. the } should not be on the base line
const status = tryCatch(() => {
        device.removeFabric(fabricIndex);
        return OperationalCertStatus.Success;
    }, FabricNotFoundError,  OperationalCertStatus.InvalidFabricIndex);

src/matter/cluster/server/OperationalCredentialsServer.ts
```diff
@@ -109,16 +110,18 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
...
-            return { status: OperationalCertStatus.InvalidFabricIndex };
-        }
+            return OperationalCertStatus.Success;
+        },
+        FabricNotFoundError, OperationalCertStatus.InvalidFabricIndex);
 
-        // TODO persist fabrics
-        // TODO: depending on cases destroy the secure session and delete all data!
+        if (status === OperationalCertStatus.Success) {
+            // TODO persist fabrics
```
mfucci, line 120: You can move this after device.removeFabric(fabricIndex);
No need to had a separate if block

mfucci (changes requested):

src/error/MatterError.ts
Apollon77: Moved to a new FIle in interaction dir because I expect that it will be more later

src/error/MatterError.ts
Apollon77: Needed to be able to throw  "the exception that matches to the statusresponse coming in ... will move ro yur approach

Apollon77: But ok you are right we can also make it very generic then we just thor an StatusResponse exception with the statuscode ... so catching needs to check instance and then the code ... also works. will adjust

src/error/MatterError.ts
Apollon77: In the concept here a return of undefined by the method would throw the error instead of using it as fallback value, so by intent

src/error/MatterError.ts
Apollon77: I renamed to fallbackValueOrFunction :-)

Apollon77:
@mfucci Adjusted and rewritten basically

src/error/MatterError.ts
Apollon77: I also added JSDoc

src/matter/interaction/InteractionErrors.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @license
+ * Copyright 2022 The node-matter Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {MatterError} from "../../error/MatterError";
+import {StatusCode} from "./InteractionMessages";
+
+/** Specific Error for when a fabric is not found. */
+export class FabricNotFoundError extends MatterError {}
```
mfucci, line 11: You should move this in FabricManager.ts since this is thrown by it and might be handled in other code path unrelated to the InteractionProtocol.

src/matter/interaction/InteractionErrors.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/** Error base Class for all errors related to the status response messages. */
+export class StatusResponseError extends MatterError {
```
mfucci, line 14: This doesn't seem to be used anywhere...
-> probably better to add when it will be needed
-> probably should be part of InteractionMessenger.ts since is the code that will handle it but difficult to be sure without seeing an actual use case

test/error/ErrorTest.ts
```diff
@@ -0,0 +1,185 @@
...
+            const error = tryCatch(() => {
+                throw new SubMatterError("test");
+            },
+            SubMatterError, "caught");
+
+            assert.equal(error, "caught");
+        });
+
+        it("tryCatch with unexpected error, throw error", () => {
+            try {
```
mfucci, line 39: It is better to unit test exception throwing with assert.throws:

```
const test = () => {
        const error = tryCatch(() => {
                    throw new Error("test");
                },
                SubMatterError, "caught");
};
assert.throws(test, new Error("test"));
```

Same in the rest of the class

test/error/ErrorTest.ts
```diff
@@ -0,0 +1,185 @@
...
+
+class SubMatterError extends MatterError {}
+class SubSubMatterError extends SubMatterError {}
+class OtherMatterError extends MatterError {}
+
+describe("Errors", () => {
+
+    context("Test tryCatch method", () => {
+        it("tryCatch without error return value", () => {
+            const error = tryCatch(() => {
```
mfucci, line 21: There is some indentation issues here that makes this code block hard to read.
Current:
```
const error = tryCatch(() => {
    throw new SubMatterError("test");
},
SubMatterError, "caught");
```

Suggested:
```
const error = tryCatch(() => {
        throw new SubMatterError("test");
    },
    SubMatterError, "caught"
);
```

(same in the rest of the class)

test/error/ErrorTest.ts
```diff
@@ -0,0 +1,185 @@
...
+        it("tryCatch with inherited error also return fallback when checking for parent error", () => {
+            const error = tryCatch(() => {
+                throw new SubSubMatterError("test");
+            },
+            SubMatterError, "caught");
+
+            assert.equal(error, "caught");
+        });
+
+        it("tryCatch with inherited error process error in handler function return dynamic fallback value", () => {
```
mfucci, line 70: You should also test what happens when the error handler throws an error...

test/error/ErrorTest.ts
```diff
@@ -0,0 +1,185 @@
...
+            SubMatterError, (error) => {
+                if (error instanceof SubSubMatterError) {
+                    return "caught";
+                }
+            });
+
+            assert.equal(error, "caught");
+        });
+
+        it("tryCatch with inherited error process error in handler function that not return fallback value throws the error", () => {
```
mfucci, line 83: This test case won't be necessary if the error handler can't return undefined.

mfucci (changes requested):

src/error/TryCatchHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Try to execute the code block and catch the error if it is of the given type.
+ * If the error is of the given type, the fallback value or the result of the function is returned.
+ * If the function returns undefined or the error type do not match, the error is normally thrown and not handled.
+ *
+ * @param codeBlock Code block to execute
+ * @param errorType Errortype to catch and handle
+ * @param fallbackValueOrFunction Fallback value or function to compute the fallback value
+ */
+export function tryCatch<T>(codeBlock: () => T, errorType: {new (...args: any[]): Error}, fallbackValueOrFunction: ErrorHandler<T> | T): T {
```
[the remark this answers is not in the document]
Apollon77: Done

src/error/TryCatchHandler.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2022 The node-matter Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+type ErrorHandler<T> = (error: Error) => T | undefined;
```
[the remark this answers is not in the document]
Apollon77: Ok

src/error/TryCatchHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Try to execute the code block and catch the error if it is of the given type.
+ * If the error is of the given type, the fallback value or the result of the function is returned.
+ * If the function returns undefined or the error type do not match, the error is normally thrown and not handled.
+ *
+ * @param codeBlock Code block to execute
+ * @param errorType Errortype to catch and handle
+ * @param fallbackValueOrFunction Fallback value or function to compute the fallback value
+ */
+export function tryCatch<T>(codeBlock: () => T, errorType: {new (...args: any[]): Error}, fallbackValueOrFunction: ErrorHandler<T> | T): T {
+    try {
```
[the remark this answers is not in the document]
Apollon77: I really like the simplicity of not having to define a function if you have a "static rturn value" as you proposed earlier ... So the function was just because there are also other cases. So I would live the complexity her ein this one place to have more simple code in all other places ;-)

src/error/MatterError.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @license
+ * Copyright 2022 The node-matter Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/** Error base class for all errors thrown by this library. */
+export class MatterError extends Error {}
+
+export function isMatterError(e: unknown): e is MatterError {
```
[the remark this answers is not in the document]
Apollon77: removed, can be added if we need it

src/matter/interaction/InteractionErrors.ts
Apollon77: I need it in 2-3 follow up PRs so I decided to already add it here to unblock next stages ... else I have another PR chain ...

It will later be thrown (e.g.) here:

Bildschirm­foto 2023-02-05 um 16 20 57

And catched in cases like ...
Bildschirm­foto 2023-02-05 um 16 22 16

Apollon77:
@mfucci Adjusted

src/error/MatterJsError.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2022 The node-matter Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Used to identify errors cases from this library.
+ */
+export enum MatterJsErrorCodes {
```
mfucci: nit: const?

mfucci (approved):

src/error/MatterJsError.ts
Apollon77: outdated

PR #214, opened 2023-02-01
Temperature measurement cluster

JimBuzbee:
First cut for adding these simple sensor clusters


Apollon77:
Hi @JimBuzbee I would like to not add all these "scripts" just for different sensor types ... So I would just go with the clusters. I would more enhance the Device to allow to set a type and then have one because 90% of the code is the same for all these types ... lets move that to a bit later ...  ok?

I will check the clusters later today

src/matter/cluster/TemperatureMeasurementCluster.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const TemperatureMeasurementCluster = Cluster({
+    id: 0x0402,
+    name: "TemperatureMeasurement",
+    revision: 4,
+    features: { }, // no features
+
+    /** @see {@link MatterApplicationClusterSpecificationV1_0} § 2.3.4 */
+    attributes: {
+       measuredValue: Attribute(0, TlvInt16.bound({ min: -27315, max: 32767 }), { default: 1 }),
```
Apollon77, line 23: Please see IlluminanceMeasurement ... I adjusted anything which was wrong there ... in fact basically:
* Quality "X" means to add a "tlvNullable around the type
* when having "semantic/dyna,ic ranges defined by other settings then I would not define them, unless they can provide an actual min or max value
* "O" in column Conformance means Optiona so use OptionalAttribute
* Please use default values on if they are really also defined
* Adding documentation per Attribute would also be cool especially relevant I thinkfor the measuredValue because is special

src/matter/cluster/WaterContentMeasurementCluster.ts
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Attribute, Cluster, Command, TlvNoArguments, TlvNoResponse } from "./Cluster";
+import { BitFlag, MatterApplicationClusterSpecificationV1_0, TlvBitmap, TlvInt16, TlvUInt16, TlvEnum, TlvField, TlvNullable, TlvObject, TlvSchema, TlvUInt8 } from "@project-chip/matter.js";
+
+/**
+ * Attributes and commands for WaterContent Measurement.
+ *
+ * @see {@link MatterApplicationClusterSpecificationV1_0} § 2.3
```
Apollon77, line 13: be carefulk when copy/paste ... it is 2.6 :-) also the other spec references are wrong.

src/matter/cluster/WaterContentMeasurementCluster.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const WaterContentMeasurementCluster = Cluster({
+    id: 0x0405,
+    name: "WaterContentMeasurement",
+    revision: 4,
```
Apollon77, line 18: revision = 3

src/matter/cluster/WaterContentMeasurementCluster.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Attribute, Cluster, Command, TlvNoArguments, TlvNoResponse } from "./Cluster";
+import { BitFlag, MatterApplicationClusterSpecificationV1_0, TlvBitmap, TlvInt16, TlvUInt16, TlvEnum, TlvField, TlvNullable, TlvObject, TlvSchema, TlvUInt8 } from "@project-chip/matter.js";
+
+/**
+ * Attributes and commands for WaterContent Measurement.
+ *
+ * @see {@link MatterApplicationClusterSpecificationV1_0} § 2.3
+ */
+export const WaterContentMeasurementCluster = Cluster({
+    id: 0x0405,
```
Apollon77, line 16: in fact we need 3 clusters with that ... because there are 3 IDs ... structure wise should be like LabelCluster maybe?

Apollon77 (changes requested):
Please look into my comments ... If you have questions just ask

src/matter/cluster/TemperatureMeasurementCluster.ts
JimBuzbee: Thanks !- Just starting to make sense of the standard

JimBuzbee:
Thanks for the comments and corrections. Agree with all. 

With regard to the multiple, nearly identical "scripts" - yeah, I was coming to the same conclusion. Maybe enhance the single script to have a command line argument to specify which device to instantiate? Command line could be messy as well. Not sure.

Now that the new clusters are working on Google Home I will play with it today to display my live weather data which for me, arrives via broadcast UDP.

src/matter/cluster/TemperatureMeasurementCluster.ts
JimBuzbee: Will work a bit today to clean up WaterContent and make sure others are correct

Apollon77:
@JimBuzbee 

> With regard to the multiple, nearly identical "scripts" - yeah, I was coming to the same conclusion. Maybe enhance the single script to have a command line argument to specify which device to instantiate? Command line could be messy as well. Not sure.

I have an older branch where I also made lightbulb vs plug-socket types configurable via commandline ... maybe we enhance this. but I went back to some groundworks. The next updates could be interesting steps forward ;-)

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
...
+import { ByteArray } from "@project-chip/matter.js";
+import { CommissionningFlowType, DiscoveryCapabilitiesSchema, ManualPairingCodeCodec, QrPairingCodeCodec } from "./codec/PairingCode.js";
+import { QrCode } from "./codec/QrCode.js";
+import { NetworkCommissioningCluster, NetworkCommissioningStatus } from "./matter/cluster/NetworkCommissioningCluster";
+import { AdminCommissioningCluster, CommissioningWindowStatus } from "./matter/cluster/AdminCommissioningCluster";
+import { AdminCommissioningHandler } from "./matter/cluster/server/AdminCommissioningServer";
+import { NetworkCommissioningHandler } from "./matter/cluster/server/NetworkCommissioningServer";
+import { FabricIndex } from "./matter/common/FabricIndex";
+
+// From Chip-Test-DAC-FFF1-8000-0007-Key.der
```
mfucci, line 49: Would be good to extract the 4 constants below in its own file and reuse it.

Yes, those certificates should autogenerated insteda of using CHIP example certificates but the code to autogenerate them is not written yet (should not be **too** complicated since this is standard crypto).

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
...
+        logger.info(`node-matter@${packageJson.version}`);
+
+        const deviceName = "Matter Sensor device";
+        const deviceType = DEVICE.HUMIDITY_SENSOR.code ;
+        const vendorName = "node-matter";
+        const passcode = 20202021;
+        const discriminator = 3840;
+        // product name / id and vendor id should match what is in the device certificate
+        const vendorId = new VendorId(0xFFF1);
+        const productName = "Matter Test DAC 0007";
```
mfucci, line 76: This is the product name written in the certificate but I think nobody checks if this matches... Maybe you can try to "Temperature Sensor" ?

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env node
```
mfucci, line 1: Can you move those files to an examples folder?
I think this will be the best place for them.

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * @license
+ * Copyright 2022 The node-matter Authors
```
mfucci, line 5: Can you add your name in CLA.md to indicate that you accept the "Contributions License Agreement"?
This is a standard file to ensure copyrights are assigned to "node-matter Authors" and that you won't commit code copyrighted by someone else.

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        // Barebone implementation of the WaterContentMeasurement cluster
+        const waterContentMeasurementClusterServer = new ClusterServer(
+            WaterContentMeasurementCluster,
+            { measuredVAlue: false },
+            { measuredValue: 1, minMeasuredValue: 1, maxMeasuredValue: 1, tolerance: 1 },
+            ({}) // WaterContentMeasurementClusterHandler()
+        );
+
+        // for testing: -watercontent "echo \$RANDOM % 10000 | bc"
```
mfucci, line 87: Maybe you can add that in a README file?
Pretty buried in the code so somebody looking to quickly hack a sensor will have trouble finding it...

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
...
+        const waterContentScript = getParameter( "watercontent" ) ?? "";
+
+        // if we have a script to check 
+        if ( waterContentScript ) {
+           function waterContentIntervalCheck() {
+              var waterContent : number  = parseInt(execSync(waterContentScript).toString().slice(0, -1)) | 0 ;
```
mfucci, line 93: Maybe you can modify commandExecutor() method in CommandLine to return the value instead and use it here?
Or add a new method: getIntCommandResult()

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
...
+        const discriminator = 3840;
+        // product name / id and vendor id should match what is in the device certificate
+        const vendorId = new VendorId(0xFFF1);
+        const productName = "Matter Test DAC 0007";
+        const productId = 0X8000;
+
+        // Barebone implementation of the WaterContentMeasurement cluster
+        const waterContentMeasurementClusterServer = new ClusterServer(
+            WaterContentMeasurementCluster,
+            { measuredVAlue: false },
```
mfucci, line 82: nit: measuredVAlue -> measuredValue

src/HumiditySensorDevice.ts
```diff
@@ -0,0 +1,213 @@
...
+            { measuredValue: 1, minMeasuredValue: 1, maxMeasuredValue: 1, tolerance: 1 },
```
mfucci, line 83: measuredValue: getIntParameter( "watercontent" )  ?? 0

This would avoid the next few lines of code...

src/HumiditySensorDevice.ts
JimBuzbee: Added myself at end

src/HumiditySensorDevice.ts
JimBuzbee: Fixed

src/HumiditySensorDevice.ts
JimBuzbee: Moved certs into own file. Will update code to include

Apollon77: Maybe call the File DevCertificates.ts or such?

src/HumiditySensorDevice.ts
Apollon77: Cool

src/HumiditySensorDevice.ts
Apollon77: I agree - Alternative would be to enhance the "Device.ts" to allow to specify a -type parameter and then have just small stub files with the "which lucster to add and how to configure"??

src/HumiditySensorDevice.ts
Apollon77: maybe a README.md in example directry if we move the files to example?

src/HumiditySensorDevice.ts
Apollon77: I like the getIntCommandResult() idea tbh

JimBuzbee: Will try and get to that tomorrow. Will also need float.

Apollon77: Then do two :-)

src/certificates.ts
```diff
@@ -0,0 +1,16 @@
+
+import { ByteArray } from "@project-chip/matter.js";
```
mfucci, line 2: For consistency, this file name should start with a capital letter

src/util/CommandLine.ts
```diff
@@ -24,4 +24,14 @@ export function commandExecutor(scriptParamName: string) {
     const script = getParameter(scriptParamName);
     if (script === undefined) return undefined;
     return () => console.log(`${scriptParamName}: ${execSync(script).toString().slice(0, -1)}`);
-}
\ No newline at end of file
+}
+
+export function getFloatCommandResult(script: string) {
+    return (parseFloat(`${execSync(script).toString().slice(0, -1)}`) || 0);
```
mfucci, line 30: Extract this since this is repeated 3 times in this class

mfucci, line 30: I don't think we want to return 0 if we cannot parse the float since this will swallow issues, making them hard to debug.
Instead, throw an Error:
```
const commandResult = executeCommand(script);
const parsedResult = parseFloat(commandResult);
if (parsed === NaN) throw new Error(`Cannot parse a float from: ${commandResult}`);
```

Same below

mfucci (changes requested):
To speed up the review, I would recommend to split this MR:

1. One for CLA change
2. One for extracting Certificates
3. One for adding the new cluster definitions,
4. One for adding the cluster implementations

1 / 2 are ready to be merged, while the other ones probably needs a few iterations

Apollon77:
@JimBuzbee I you need support on this please tell ... 

@mfucci  @JimBuzbee For me I would go with adding the clusters only for now but leaving the examples out for now. I will work on the more high level api soon ... after that all that changes anyway. I could use the  examples as basis for a new approach. I think @JimBuzbee can sill use the examples for himself locally ... so all fine here.

src/certificates.ts
JimBuzbee: OK. @Apollon77 also suggested name change to DeveloperCertificates.

src/util/CommandLine.ts
JimBuzbee: ok

JimBuzbee: Agree. During debug was just concerned with getting any value, but not appropriate for production. Will change.

JimBuzbee:
Regarding @Apollon77 suggestion to myself and @mfucci that we limit this merge and leave out the examples for now- I have no problem with this.  The process is new to me - Should I close this one and open up new, individual pull requests for each component with the suggestions applied?

I also have two new clusters written and tested as much as I can in Google Home: Occupancy Sensing and Boolean State (used in contact sensor device).

Apollon77:
> Should I close this one and open up new, individual pull requests for each component with the suggestions applied?

I think this would be ideal ... maybe instead of closing convert this PR to a "draft" (right on top below "reviewers"). Draft means that it is not yet ready to be revieed, but so nooen comes to the idea and we still have it "to remember" :-)

Then open one new PR for each of the cluster definitions ... Thank you very much!

Apollon77:
@JimBuzbee To do  an own PR per cluster you basically need to create an own branch for each cluster file ... If you need more support or such you can also try to join discord on https://discord.gg/zDJfbPBf and we can iterate on it more direct

JimBuzbee:
@Apollon77 - Just tried the Boolean State Cluster PR - hope that was right!

Apollon77:
@JimBuzbee I think we have all this from here in the meantime in main branch right? Your "CLI exec" is also in the other draft PR ...

JimBuzbee:
I think we have it all in now.  I re-added my name/update to the CLA in the LevelControl PR so that is now in too.

PR #217, opened 2023-02-03
Add nonVolatile and omitChanges Attribute options

Apollon77:
This PR adds two new Attribute options to be used in later PRs
* **nonVolatile** means that the value is persisted and initialized again on reboot/restart
* **ommitChanges** means that value changes should not send in subscriptions

I decided to add them as options and not types because I think the current types (optional, writable. normal and all combination) are already enough :-)

The PR also contains some minor typo fixes

### DIFF
--- a/src/matter/cluster/Cluster.ts
+++ b/src/matter/cluster/Cluster.ts
@@ -28,16 +28,16 @@ export const enum AccessLevel {
 }
 
 /* Interfaces and helper methods to define a cluster attribute */
-export interface Attribute<T> { id: number, schema: TlvSchema<T>, optional: boolean, readAcl: AccessLevel, writable: boolean, writeAcl?: AccessLevel, default?: T }
+export interface Attribute<T> { id: number, schema: TlvSchema<T>, optional: boolean, readAcl: AccessLevel, writable: boolean, nonVolatile?: boolean, omitChanges?: boolean, writeAcl?: AccessLevel, default?: T }
 export interface OptionalAttribute<T> extends Attribute<T> { optional: true }
 export interface WritableAttribute<T> extends Attribute<T> { writable: true }
 export interface OptionalWritableAttribute<T> extends OptionalAttribute<T> { writable: true }
 export type AttributeJsType<T extends Attribute<any>> = T extends Attribute<infer JsType> ? JsType : never;
-interface AttributeOptions<T> { default?: T, readAcl?: AccessLevel, writeAcl?: AccessLevel };
-export const Attribute = <T, V extends T>(id: number, schema: TlvSchema<T>, { default: conformanceValue, readAcl = AccessLevel.View }: AttributeOptions<V> = {}): Attribute<T> => ({ id, schema, optional: false, writable: false, default: conformanceValue, readAcl });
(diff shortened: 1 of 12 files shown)


src/matter/cluster/Cluster.ts
```diff
@@ -28,16 +28,16 @@ export const enum AccessLevel {
 }
 
 /* Interfaces and helper methods to define a cluster attribute */
-export interface Attribute<T> { id: number, schema: TlvSchema<T>, optional: boolean, readAcl: AccessLevel, writable: boolean, writeAcl?: AccessLevel, default?: T }
+export interface Attribute<T> { id: number, schema: TlvSchema<T>, optional: boolean, readAcl: AccessLevel, writable: boolean, nonVolatile?: boolean, omitChanges?: boolean, writeAcl?: AccessLevel, default?: T }
```
mfucci, line 31: nonVolatile => persistent?
Better to always use positive names in flag, otherwise this adds cognitive burden, like this:
`if (!nonVolatile) return;`

:-)

Same for omitChanges => observable (or something like this)

I don't think nonVolatile / omitChanges should be optional: a default value can always defined in the declaration methods below.
writeAcl is optional because it only makes sense when the attribute is writable
default is optional since not all the attributes have a default

src/matter/cluster/OperationalCredentialsCluster.ts
```diff
@@ -242,22 +242,22 @@ export const OperationalCredentialsCluster = Cluster({
...
 
         /** Contains the number of Fabrics to which the device is currently commissioned. */
-        commissionedFabrics: Attribute(3, TlvUInt8),
+        commissionedFabrics: Attribute(3, TlvUInt8, { nonVolatile: true }),
 
         /** Contains a read-only list of Trusted Root CA Certificates installed on the Node. */
-        trustedRootCertificates: Attribute(4, TlvArray(TlvByteString, { maxLength: 400 })),
+        trustedRootCertificates: Attribute(4, TlvArray(TlvByteString, { maxLength: 400 }), { nonVolatile: true, omitChanges: true }),
 
         /** Contain accessing fabric index. */
```
mfucci, line 259: nit: Contains

mfucci (changes requested):

src/matter/cluster/Cluster.ts
Apollon77: Naming wise we are then (again) on the question how close to be With the specs when it comes to names ;-) it is officially called that way there.

But yes I will rename it then.

Apollon77: Ok for nonVolative -> persistent ... all fine.

But for omitChanges aka "do NOT send updates on subscriptions" i can not find a positive name that would not also "reverse" the meaning ... And if we go to "subscribable" as the "positive opposite" then User (and we) would need to have a big brain effort to map with the specs because this is then not "clear" (because if the specs say "omit changes" we would need to set "subscribable=false" with a default of true. Additionally according to specsit is indeed "subscribable" (aka would not rturn an error on subscribe) but would in fact never get any value, so also the name would be misleading.

Apollon77:
@mfucci Adjusted prsistent as proposed and also "defaults" to false now both. For "omitChanges" see above: I would vote for staying with this name to be more easy mappable to the specs

mfucci (approved):

PR #218, opened 2023-02-03
Enhance error handling/logging for network and message errors

Apollon77:
This PR just catches some error and make sure they are visible and provide more details.
On network level the "catching on the lowest level" should also make sure it is announced as good as possible

### DIFF
--- a/src/matter/common/ExchangeManager.ts
+++ b/src/matter/common/ExchangeManager.ts
@@ -53,7 +53,7 @@ export class ExchangeManager<ContextT> {
     addNetInterface(netInterface: NetInterface) {
         this.netListeners.push(netInterface.onData((socket, data) => {
             this.onMessage(socket, data)
-                .catch(error => logger.error(error.stack));
+                .catch(error => logger.error(error));
         }));
     }
 
--- a/src/matter/common/ExchangeManager.ts
+++ b/src/matter/common/ExchangeManager.ts
@@ -53,7 +53,7 @@ export class ExchangeManager<ContextT> {
     addNetInterface(netInterface: NetInterface) {
         this.netListeners.push(netInterface.onData((socket, data) => {
             this.onMessage(socket, data)
-                .catch(error => logger.error(error));
+                .catch(error => logger.error(error.stack));
         }));
     }
 
--- a/src/net/UdpMulticastServer.ts
+++ b/src/net/UdpMulticastServer.ts
@@ -55,7 +55,11 @@ export class UdpMulticastServer {
             const { ips } = this.network.getIpMac(netInterface) ?? { ips: [] };
             await Promise.all(ips.map(async ip => {
                 const iPv4 = isIPv4(ip);
-                await (await this.broadcastChannels.get(netInterface, iPv4)).send(iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6, this.broadcastPort, message).catch(err => logger.debug(`${netInterface}: ${err}`));
+                try {
+                    await (await this.broadcastChannels.get(netInterface, iPv4)).send(iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6, this.broadcastPort, message);
+                } catch (err) {
+                    logger.debug(`${netInterface}: ${(err as Error).message}`);
+                }
             }));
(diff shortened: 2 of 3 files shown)


src/matter/common/ExchangeManager.ts
```diff
@@ -53,7 +53,7 @@ export class ExchangeManager<ContextT> {
     addNetInterface(netInterface: NetInterface) {
         this.netListeners.push(netInterface.onData((socket, data) => {
             this.onMessage(socket, data)
-                .catch(error => logger.error(error));
+                .catch(error => logger.error(error.stack));
```
mfucci, line 56: Maybe this should be modified in Logger.ts to always display the stack on .error() when the argument is of Error class?
I don't see a reason why not to always display the stack...

src/net/UdpMulticastServer.ts
```diff
@@ -52,7 +55,7 @@ export class UdpMulticastServer {
             const { ips } = this.network.getIpMac(netInterface) ?? { ips: [] };
             await Promise.all(ips.map(async ip => {
                 const iPv4 = isIPv4(ip);
-                await (await this.broadcastChannels.get(netInterface, iPv4)).send(iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6, this.broadcastPort, message);
+                await (await this.broadcastChannels.get(netInterface, iPv4)).send(iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6, this.broadcastPort, message).catch(err => logger.debug(`${netInterface}: ${err}`));
```
mfucci, line 58: This is an async function so maybe better to write this using try / catch;
```
try {
     const { ips } = this.network.getIpMac(netInterface) ?? { ips: [] };
            await Promise.all(ips.map(async ip => {
                const iPv4 = isIPv4(ip);
                await (await this.broadcastChannels.get(netInterface, iPv4)).send(iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6, this.broadcastPort, message));
              }));
        }));
} catch (error) {
    logger.debug(`${netInterface}: ${error}`);
}
```

mfucci, line 58: logger.warn ?
I think this match the definition of a warning: error but with no expected functional impact

mfucci (changes requested):

src/net/UdpMulticastServer.ts
Apollon77: Hm ... for Me warn is something where the user should react on but here he can not even do anything. That's why for me maximal info ;-)

src/matter/common/ExchangeManager.ts
Apollon77: Ok

src/net/UdpMulticastServer.ts
Apollon77: I adjusted it to info.

Apollon77:
@mfucci Ready for re-review

src/net/UdpMulticastServer.ts
```diff
@@ -52,7 +55,11 @@ export class UdpMulticastServer {
             const { ips } = this.network.getIpMac(netInterface) ?? { ips: [] };
             await Promise.all(ips.map(async ip => {
                 const iPv4 = isIPv4(ip);
-                await (await this.broadcastChannels.get(netInterface, iPv4)).send(iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6, this.broadcastPort, message);
+                try {
+                    await (await this.broadcastChannels.get(netInterface, iPv4)).send(iPv4 ? this.broadcastAddressIpv4 : this.broadcastAddressIpv6, this.broadcastPort, message);
+                } catch (err) {
+                    logger.debug(`${netInterface}: ${(err as Error).message}`);
```
mfucci, line 61: logger.info?

mfucci (dismissed):

src/net/UdpMulticastServer.ts
Apollon77: adjusted. @mfucci Re-Review please ;-)) (or change that setting? ;-) )

mfucci (approved):

PR #219, opened 2023-02-03
Fix Subscriptions

Apollon77:
This PR is a bigger overhaul to fix issues the former codebase had with subscriptions. basically the issues are about the following topics:
1. The subscription IDs needs to be "unique for the publisher". The former codebase had then always starting at 0 for each SecureSession where a subscription was started, so they were always overlapping as soon as a device did multiple subscriptions. This lead to issues with at least iOS. The PR changes to a "started at random" globally increase ID (like chiptool does it too)
2. The subscription initialization process was not followed. We missed the "priming the subscription with initial data" which needs to happen after the subscripion Request and before sending the subscription initialization response.

The PR also adds handling to only send out data updates and not slways the full data (as it was before).

Additionally the PR also adds support for incoming chunked data reports when used as controller because also needed for testing puroposes.

I could not really split that up further because the changes are somewhat connected, so I hope we can review as is.

I tested with iOS, Google and Alexa

### DIFF
--- a/src/matter/interaction/InteractionMessenger.ts
+++ b/src/matter/interaction/InteractionMessenger.ts
@@ -109,7 +109,7 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
     async handleRequest(
         handleReadRequest: (request: ReadRequest) => DataReport,
         handleWriteRequest: (request: WriteRequest) => WriteResponse,
-        handleSubscribeRequest: (request: SubscribeRequest, messenger: InteractionServerMessenger) => Promise<SubscribeResponse | undefined>,
+        handleSubscribeRequest: (request: SubscribeRequest, messenger: InteractionServerMessenger) => Promise<void>,
         handleInvokeRequest: (request: InvokeRequest, message: Message) => Promise<InvokeResponse>,
         handleTimedRequest: (request: TimedRequest) => Promise<void>,
     ) {
@@ -130,12 +130,8 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                         break;
                     case MessageType.SubscribeRequest:
                         const subscribeRequest = TlvSubscribeRequest.decode(message.payload);
-                        const subscribeResponse = await handleSubscribeRequest(subscribeRequest, this);
-                        if (subscribeResponse === undefined) {
-                            await this.sendStatus(StatusCode.Success);
-                        } else {
-                            await this.exchange.send(MessageType.SubscribeResponse, TlvSubscribeResponse.encode(subscribeResponse));
-                        }
+                        await handleSubscribeRequest(subscribeRequest, this);
+                        // response is sent by handler
                         break;
                     case MessageType.InvokeCommandRequest:
                         const invokeRequest = TlvInvokeRequest.decode(message.payload);
@@ -196,6 +192,10 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
 
         await this.exchange.send(MessageType.ReportData, TlvDataReport.encode(dataReport));
     }
+
+    async send(messageType: number, payload: ByteArray) {
+        return this.exchange.send(messageType, payload);
+    }
 }
 
 export class InteractionClientMessenger extends InteractionMessenger<MatterController> {
@@ -211,39 +211,20 @@ export class InteractionClientMessenger extends InteractionMessenger<MatterContr
 
     async sendSubscribeRequest(subscribeRequest: SubscribeRequest) {
         await this.exchange.send(MessageType.SubscribeRequest, TlvSubscribeRequest.encode(subscribeRequest));
-        const initialSubscriptionValues = await this.readDataReport();
-        if (!initialSubscriptionValues.values) initialSubscriptionValues.values = [];
-        const subscriptionId = initialSubscriptionValues.subscriptionId;
-        if (subscriptionId === undefined) {
-            await this.sendStatus(StatusCode.Failure);
-            throw new Error('Subscription ID not provided');
-        }
-        await this.sendStatus(StatusCode.Success);
-        if (initialSubscriptionValues.moreChunkedMessages) {
-            while (true) {
-                const dataReport = await this.readDataReport();
-                if (dataReport.subscriptionId !== subscriptionId) {
-                    await this.sendStatus(StatusCode.Failure);
-                    throw new Error(`Received subscription ID ${dataReport.subscriptionId} instead of ${subscriptionId}`);
-                }
-                await this.sendStatus(StatusCode.Success);
-                if (dataReport.values && dataReport.values.length > 0) {
-                    initialSubscriptionValues.values.push(...dataReport.values);
-                }
-                if (!dataReport.moreChunkedMessages) {
-                    initialSubscriptionValues.moreChunkedMessages = false;
-                    break;
-                }
-            }
-        }
+
+        const report = await this.readDataReport();
+        const { subscriptionId } = report;
+
         const subscribeResponseMessage = await this.nextMessage(MessageType.SubscribeResponse);
         const subscribeResponse = TlvSubscribeResponse.decode(subscribeResponseMessage.payload);
+
         if (subscribeResponse.subscriptionId !== subscriptionId) {
             throw new Error(`Received subscription ID ${subscribeResponse.subscriptionId} instead of ${subscriptionId}`);
         }
+
         return {
             subscribeResponse,
-            initialSubscriptionValues,
+            report,
         };
     }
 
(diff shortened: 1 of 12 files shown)


mfucci:
You should probably walk me through this one...

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -206,8 +209,42 @@ export class InteractionClientMessenger extends InteractionMessenger<MatterContr
...
 
-    sendSubscribeRequest(subscribeRequest: SubscribeRequest) {
-        return this.request(MessageType.SubscribeRequest, TlvSubscribeRequest, MessageType.SubscribeResponse, TlvSubscribeResponse, subscribeRequest);
+    async sendSubscribeRequest(subscribeRequest: SubscribeRequest) {
+        await this.exchange.send(MessageType.SubscribeRequest, TlvSubscribeRequest.encode(subscribeRequest));
+        const initialSubscriptionValues = await this.readDataReport();
+        if (!initialSubscriptionValues.values) initialSubscriptionValues.values = [];
+        const subscriptionId = initialSubscriptionValues.subscriptionId;
+        if (subscriptionId === undefined) {
+            await this.sendStatus(StatusCode.Failure);
```
mfucci, line 218: This should be done by the catch-all method

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -206,8 +209,42 @@ export class InteractionClientMessenger extends InteractionMessenger<MatterContr
         return this.request(MessageType.ReadRequest, TlvReadRequest, MessageType.ReportData, TlvDataReport, readRequest);
     }
 
-    sendSubscribeRequest(subscribeRequest: SubscribeRequest) {
-        return this.request(MessageType.SubscribeRequest, TlvSubscribeRequest, MessageType.SubscribeResponse, TlvSubscribeResponse, subscribeRequest);
+    async sendSubscribeRequest(subscribeRequest: SubscribeRequest) {
+        await this.exchange.send(MessageType.SubscribeRequest, TlvSubscribeRequest.encode(subscribeRequest));
+        const initialSubscriptionValues = await this.readDataReport();
```
mfucci, line 214: Move chunking into readDataReport

src/matter/interaction/InteractionServer.ts
```diff
@@ -224,30 +226,49 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
...
-    handleSubscribeRequest(exchange: MessageExchange<MatterDevice>, { minIntervalFloorSeconds, maxIntervalCeilingSeconds, attributeRequests, keepSubscriptions }: SubscribeRequest): SubscribeResponse | undefined {
+    async handleSubscribeRequest(exchange: MessageExchange<MatterDevice>, { minIntervalFloorSeconds, maxIntervalCeilingSeconds, attributeRequests, eventRequests, keepSubscriptions }: SubscribeRequest, messenger: InteractionServerMessenger): Promise<SubscribeResponse | undefined> {
         logger.debug(`Received subscribe request from ${exchange.channel.getName()}`);
 
         if (!exchange.session.isSecure()) throw new Error("Subscriptions are only implemented on secure sessions");
         const session = exchange.session as SecureSession<MatterDevice>;
         const fabric = session.getFabric();
         if (fabric === undefined) throw new Error("Subscriptions are only implemented after a fabric has been assigned");
 
+        if ((!attributeRequests || attributeRequests.length === 0) && (!eventRequests || eventRequests.length === 0)) {
```
mfucci, line 237: attributeRequests !== undefined

src/matter/interaction/InteractionServer.ts
```diff
@@ -224,30 +226,49 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
...
+            await messenger.sendStatus(StatusCode.InvalidAction);
```
mfucci, line 238: You can throw a typed error instead

src/matter/interaction/InteractionServer.ts
```diff
@@ -93,14 +95,13 @@ function toHex(value: number | undefined) {
...
 
-const logger = Logger.get("InteractionProtocol");
-
 export class InteractionServer implements ProtocolHandler<MatterDevice> {
     private readonly endpoints = new Map<number, { name: string, code: number, clusters: Map<number, ClusterServer<any, any, any, any>> }>();
     private readonly attributes = new Map<string, AttributeServer<any>>();
     private readonly attributePaths = new Array<AttributePath>();
     private readonly commands = new Map<string, CommandServer<any, any>>();
     private readonly commandPaths = new Array<CommandPath>();
+    private nextSubscriptionId = Math.floor(Math.random() * 0xFFFFFFFF);
```
mfucci, line 104: Use Crypto library instead

src/matter/interaction/InteractionServer.ts
```diff
@@ -224,30 +226,49 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
...
             if (maxIntervalCeilingSeconds < minIntervalFloorSeconds) throw new Error("maxIntervalCeilingSeconds should be greater or equal to minIntervalFloorSeconds");
 
-            const subscriptionId = session.addSubscription(SubscriptionHandler.Builder(session.getContext(), fabric, session.getPeerNodeId(), attributes, minIntervalFloorSeconds, maxIntervalCeilingSeconds));
+            let attributes = this.getAttributes(attributeRequests);
+
+            // TODO: Interpret specs:
+            // The publisher SHALL compute an appropriate value for the MaxInterval field in the action. This SHALL respect the following constraint: MinIntervalFloor ≤ MaxInterval ≤ MAX(SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT=60mn, MaxIntervalCeiling)
+
+            if (this.nextSubscriptionId === 0xFFFFFFFF) this.nextSubscriptionId = 0;
+            const subscriptionHandler = new SubscriptionHandler(this.nextSubscriptionId++, session.getContext(), fabric, session.getPeerNodeId(), attributes, minIntervalFloorSeconds * 1000, maxIntervalCeilingSeconds * 1000);
```
mfucci, line 261: Maybe * 1000 can be moved inside the class

src/matter/interaction/InteractionServer.ts
```diff
@@ -224,30 +226,49 @@ export class InteractionServer implements ProtocolHandler<MatterDevice> {
...
+            session.addSubscription(subscriptionHandler);
+            const subscriptionId = subscriptionHandler.subscriptionId;
```
mfucci, line 263: Use const subscriptionId = this.nextSubscriptionId++, to make this m,ore readable

src/matter/interaction/SubscriptionHandler.ts
```diff
@@ -7,27 +7,32 @@
...
+import { AttributeWithPath, AttributePath, INTERACTION_PROTOCOL_ID, attributePathToId } from "./InteractionServer";
 import { Time, Timer } from "../../time/Time";
 import { NodeId } from "../common/NodeId";
 import { TlvSchema } from "@project-chip/matter.js";
 import { Logger } from "../../log/Logger";
+import { SecureSession } from "../session/SecureSession";
 
 const logger = Logger.get("SubscriptionHandler");
 
+const SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_MS = 1000 * 60 * 60;
```
mfucci, line 19: Add /* 1 hour */ comment

src/matter/interaction/SubscriptionHandler.ts
```diff
@@ -7,27 +7,32 @@
...
 }
 
 export class SubscriptionHandler {
-
-    static Builder = (server: MatterDevice, fabric: Fabric, peerNodeId: NodeId, attributes: AttributeWithPath[], minIntervalFloorSeconds: number, maxIntervalCeilingSeconds: number) => (subscriptionId: number) => new SubscriptionHandler(subscriptionId, server, fabric, peerNodeId, attributes, minIntervalFloorSeconds * 1000, maxIntervalCeilingSeconds * 1000);
-
     private lastUpdateTimeMs = 0;
-    private readonly listener = () => this.sendUpdate();
-    private updateTimer: Timer;
+    private updateTimer: Timer | null = null;
```
mfucci, line 30: Can be always defined

src/matter/interaction/SubscriptionHandler.ts
```diff
@@ -38,49 +43,119 @@ export class SubscriptionHandler {
...
+                    version,
+                    value: schema.encodeTlv(value),
+                },
+            })),
+        });
+
+        await messenger.waitForSuccess();
+
+        this.isActive = true;
+        this.updateTimer = Time.getTimer(this.sendInterval, () => this.sendUpdate()).start();
```
mfucci, line 115: You should set  this.lastUpdateTimeMs

src/matter/interaction/SubscriptionHandler.ts
```diff
@@ -38,49 +43,119 @@ export class SubscriptionHandler {
...
+                    path,
+                    version,
+                    value: schema.encodeTlv(value),
+                },
+            })),
+        });
+
+        await messenger.waitForSuccess();
+
+        this.isActive = true;
```
mfucci, line 114: Add an activate() method after SUbribeRespomse is sent

src/matter/interaction/InteractionMessenger.ts
Apollon77: there is no catch all for client usage right now, so stays as is

src/matter/interaction/InteractionServer.ts
Apollon77: Ok we throw but will just work when the other PR is accepted, so now returns Failure, which is also okish

Apollon77:
@mfucci Adjusted to what we agreed on this morning. 
BTW: I stumbled over one case where a MessageExchange was created pot. twice because of an async case when two messages come in at the same time. I adjusted that too (and basically the fake network does that sometimes, but other story - could for delayed receive of messages also be a reals world scenario)

So, Ready for re-review

mfucci (approved):

PR #222, opened 2023-02-03
Make DataReport fields optional because clients send so

Apollon77:
According to specs the values and supporessResponse fields are mandatory in a DataReport sent by a subscription.

The reality is that chip-tool apps sometimes send the reports without these fields, so we make them optional and add Code to handle this

### DIFF
--- a/src/matter/interaction/InteractionMessages.ts
+++ b/src/matter/interaction/InteractionMessages.ts
@@ -95,9 +95,9 @@ export const TlvAttributeReport = TlvObject({
 
 export const TlvDataReport = TlvObject({
     subscriptionId: TlvOptionalField(0, TlvUInt32),
-    values: TlvField(1, TlvArray(TlvAttributeReport)),
+    values: TlvOptionalField(1, TlvArray(TlvAttributeReport)), // Mandatory but chip tool do not send it
     moreChunkedMessages: TlvOptionalField(3, TlvBoolean),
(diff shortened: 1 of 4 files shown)


src/matter/interaction/InteractionMessages.ts
```diff
@@ -95,9 +95,9 @@ export const TlvAttributeReport = TlvObject({
 
 export const TlvDataReport = TlvObject({
     subscriptionId: TlvOptionalField(0, TlvUInt32),
-    values: TlvField(1, TlvArray(TlvAttributeReport)),
+    values: TlvOptionalField(1, TlvArray(TlvAttributeReport)), // Mandatory but chip tool do not send it
```
mfucci, line 98: I was wondering if we should not instead add support for a "default value" at the TLV level.
1. This is already used at a couple of places and we cannot represent it (on commands).
2. This will make keep the code simple for the specific bug addressed here.

TlvField(index: number, schema: TlvSchema<T>, default?: T)
When we decode the TLV object, is the field is missing, we use "default" value instead.

Should be an easy change

Apollon77: Good idea. I would add a todo then here and do a pr in matter.js because we need it from there together with a release.

Apollon77:
Converted this to draft in favor of the fallback PR ... Once merged and releaed there I will update/redo this PR

Apollon77:
superceeded by #263

PR #226, opened 2023-02-04
Add minimal persistence classes

Apollon77:
This is a first minimalistic Persistence implementation. It uses a Map internally with a "context" key structure and allows to be stored a JSON and one implementation saves the data using Nodejs FS, but all splitted up so that a Browser implementation could "just" store the JSON differently.

The storage also preserves bigint, Buffer and Uint8Arrays and restores them correctly.


mfucci:
Thanks for writing this!

The APIs to use it seems a little complex though.
I suggest to reorganize the methods to make it simple to use in the code, like this:

const persistentStore = Persistence.get().getStore("my_store");

function myFunction() {
    const myPersistentValue = persistentStore.get<number>("myValue");
    myPersistentValue++;
    persistentStore.set(myPersistentValue);
}

Somewhere this method should be called:
await Persistence.get().init();

Persistence.init(): loads all the persistence in memory (async)
Persistence.getStore(key: string): returns a store to isolate data from one context to another and avoid data name collision
Store.get<T>(): retrieves the data from the memory
Store.set<T>(value: T): sets the data in memory and triggers a background task to persist them

WDYT?

Apollon77:
Hi, i agree that to "use" the storage a singleton approach is more easy and I can add this easiely. 

Internally I would like to keep such a structure because I want to be flexible with how data is actually really stored - like SQLite, redis, jsonl-db and such and all would be different. Alone in iobroker as example we usually use jsonl-db (https://github.com/AlCalzone/jsonl-db) or use an even higher abstraction level ... I want to allow this. 

So with the singleton idea internally Persistence.init(KeyValueStoreInstance) would be the only difference to your approach to set the real storage instance which is then used.

Ok?

Apollon77:
@mfucci Ok, I now added the Singleton approach. I also adjusted my local version which uses all of this and all is fine that way ... only issues exists with the signleton approach is in testing because with Singleton we in fact can only have one persistance object (because singleton) for both Controller and Server usage. Normally only one exists in a process and so it is no issue, but for testing we use both in the same process.

The overlapping can happen currently in SessionManager when it is about remembering resumptionData because this is used by both implementations - no idea if we later have even more cases.

Apollon77:
@mfucci seems we also missed the approval here too ... changes done as agreed, Singleton removed again.

Apollon77:
@mfucci Ok did some further adjustments to your proposal out of "usage requirements"
1.) PersistenceManager.get can be called without default value but then throws an error if value do not exist. Sometimes you can not provide a defautl value (e.g. if it is an Object or such). To support this I added a "has" method to check if a certain key is set ... so you can check before and make sure there is a value
2.) Storage got a "close" method to "store" the data at the end
3.) The storage file not existing is now catched and allowed to get an empty storage
4.) I adjusted the storage of our special types and decided to use the "return as JSON string so it do not get processed in general, but double encoded, so we are sure that no stringify/parse is adjusting our value It is then parsed. If you more like the "normal object way" then we can also change it to return an actual object structure.

src/persistence/StorageNode.ts
```diff
@@ -0,0 +1,102 @@
...
+    private allowToCommit = false;
+
+    constructor(
+        private readonly path: string,
+    ) {
+        super();
+    }
+
+    async initialize() {
+        if (Object.keys(this.store).length > 0) {
```
mfucci, line 31: initialize() should be called before this object is used, so you can simply use a boolean initialized to track this and prevent get set calls.
No need for the allowToCommit as well.

src/persistence/StorageNode.ts
```diff
@@ -0,0 +1,102 @@
...
+            throw new Error("Storage contains already values, can not initialize!");
+        }
+        try {
```
mfucci, line 34: The code will be more readable if you make an explicit check if the file exists...
I guess this is what the try / catch is doing but it is not clear.

src/persistence/StorageNode.ts
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+        this.allowToCommit = true;
+    }
+
+    set<T>(context: string, key: string, value: T): void {
+        super.set(context, key, value);
+        if (!this.waitForCommit) {
+            this.waitForCommit = true;
+            this.commitTimer.start();
```
mfucci, line 48: I think waitForCommit is equivalent to checking if the timer is running or not.
So it would be better to add a new API isRunning() on Timer, this will simplify this code.

src/persistence/StorageNode.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    private async commit() {
+        if (!this.allowToCommit) return;
+        this.waitForCommit = false;
+        try {
+            await writeFile(this.path, this.toJson(this.store), "utf-8");
+        } catch (error) {
+            logger.error("Failed to write storage file", error);
```
mfucci, line 58: Do we really want to catch and continue this error?
Might be fine to just let it crash the device, since things are not working as expected.

src/persistence/StorageNode.ts
Apollon77: The allowToCommit is to prevent changes after "close" is called ... so has nothing really to do with initialization, but yes could be combined. I now combined it into initilized flag ... If you have a better name just tell, because "close" now also resets "initialized" to false

src/persistence/StorageNode.ts
Apollon77: The best practice way also by Node.js official documentation is to do it that way, because exists (even the Sync variant) is kind of async when it comes to program flow, so there can be edge cases.

see https://nodejs.org/api/fs.html#fsexistspath-callback

> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile(), or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

src/persistence/StorageNode.ts
Apollon77: Yes I have this on. my todo list but do not want to add that here and also not want to add another dependent PR for now. Please lets postpone this. I add a TODO

src/persistence/StorageNode.ts
Apollon77: The problem is that this error would be not catchable at all because triggered from within a timer, but Ok lets do that for now and see if we get issues with it

src/persistence/StorageNode.ts
mfucci: Better not to reset the initialized flag.
Usually the flag to track if the object is closed is named "closed".

Note: close() is never called in the current implementation...

src/persistence/StorageNode.ts
mfucci: OK.

Since this logic is non-obvious, better to move it to an util class so it can be done consistently everywhere and doesn't increase the complexity of the code here.

For instance called FileSystem.ts. On it, create a method read that returns undefined if the file doesn't exist and add you reply above as a comment in the code

mfucci (approved):

src/persistence/StorageNode.ts
Apollon77: I would not do that now because we onkly have this one place where we read a file

src/persistence/StorageNode.ts
Apollon77: Ok I added another closed flag now. and close on PeristenceManager

Issue #227, opened 2023-02-06
Implement the Boolean State Cluster


JimBuzbee:
As specified in the Matter Application Cluster Specification Version 1.0, section 1.7

Apollon77:
fixed in #229

PR #229, opened 2023-02-06
Initial commit of Boolean State Cluster as per issue #227

JimBuzbee:
Hope I'm doing this right!

### DIFF
--- a/src/matter/cluster/BooleanStateCluster.ts
+++ b/src/matter/cluster/BooleanStateCluster.ts
@@ -8,7 +8,7 @@ import { Attribute, OptionalEvent, EventPriority, OptionalAttribute, Cluster, Co
 import { TlvField, TlvBoolean, MatterApplicationClusterSpecificationV1_0 } from "@project-chip/matter.js";
 
 /**
- * Attributes and commands for Illuminance Measurement.
+ * Attributes and commands for Boolean State.
  *
  * @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7
  */
@@ -16,7 +16,6 @@ export const BooleanStateCluster = Cluster({
     id: 0x0045,
     name: "BooleanState",
     revision: 1,
-    features: { }, // no features
 
     /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7.5 */
     attributes: {
@@ -29,7 +28,4 @@ export const BooleanStateCluster = Cluster({
         /** generated when the StateValue attribute changes */
         stateChange: OptionalEvent(0, EventPriority.Info, { stateValue: TlvField(0, TlvBoolean)  }),
     },
-
-    /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7.6 */
-    commands: { },
 });
--- a/src/matter/cluster/BooleanStateCluster.ts
+++ b/src/matter/cluster/BooleanStateCluster.ts
@@ -19,8 +19,10 @@ export const BooleanStateCluster = Cluster({
 
     /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7.4 */
     attributes: {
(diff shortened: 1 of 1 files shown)


src/matter/cluster/BooleanStateCluster.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Attributes and commands for Illuminance Measurement.
+ *
+ * @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7
+ */
+export const BooleanStateCluster = Cluster({
+    id: 0x0045,
+    name: "BooleanState",
+    revision: 1,
+    features: { }, // no features
```
mfucci, line 19: nit: features is optional so you don't need to define this parameter is there are no features

src/matter/cluster/BooleanStateCluster.ts
```diff
@@ -0,0 +1,35 @@
...
+        * For example, in a Contact Sensor device type, FALSE=open or no contact, TRUE=closed or contact */
+       stateValue: Attribute(0x00, TlvBoolean),
+    },
+        /** @see {@link MatterApplicationClusterSpecificationV1_0 1.7.5} */
+    events: {
+        /** generated when the StateValue attribute changes */
+        stateChange: OptionalEvent(0, EventPriority.Info, { stateValue: TlvField(0, TlvBoolean)  }),
+    },
+
+    /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7.6 */
```
mfucci, line 33: This section doesn't exist in the specs...
You can delete "commands: { }," if there are no commands, this is an optional parameter

mfucci (dismissed):

src/matter/cluster/BooleanStateCluster.ts
JimBuzbee: Thanks - Fixed and pushed

src/matter/cluster/BooleanStateCluster.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Attributes and commands for Boolean State.
+ *
+ * @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7
+ */
+export const BooleanStateCluster = Cluster({
+    id: 0x0045,
+    name: "BooleanState",
+    revision: 1,
+
+    /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7.5 */
```
Apollon77, line 20:
```suggestion
    /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.7.4 */
```

src/matter/cluster/BooleanStateCluster.ts
```diff
@@ -0,0 +1,31 @@
...
+    attributes: {
+       /** The semantics of this boolean state are defined by the device type using this cluster.
+        * For example, in a Contact Sensor device type, FALSE=open or no contact, TRUE=closed or contact */
```
Apollon77, line 23: If a one liner is not posible then lets format the comment correctly

```suggestion
       /** 
        * The semantics of this boolean state are defined by the device type using this cluster.
        * For example, in a Contact Sensor device type, FALSE=open or no contact, TRUE=closed or contact 
        */
```

Apollon77 (changes requested):
Last nits :-) You can just accept the suggetsions. Then I think we can merge. Thank you!

JimBuzbee:
Will start pushing other Clusters today. Forgot to mention, tested this Cluster on Google Home using the "Contact Sensor" device. Seems to work fine

Apollon77 (approved):

Apollon77:
Thank you and welcome to the contributors :-)

JimBuzbee:
Thanks for putting up with my fumbling :-) Before this, I'd never written a line of Typescript and am still a git novice, but I'll do what I can.

PR #237, opened 2023-02-11
Implement retransmission backoff as per spec and reject with error

Apollon77:
This PR is implementing a retransmission as per spec, especially regarding
* the increasing delay fir backoff time
* default number of tries (5 instead of 2)
* respecing different backoffs based on node active status of session

Additionally there was no promise rejection in the case of "all retransmissions were not answered" so there was a good chance to have just a haning code that never continues when waiting for a send result. So now also this case rejects (yes might need some error handling on other code later) and we also reject with an error object instead of nothing :-)

### DIFF
--- a/src/matter/common/MessageExchange.ts
+++ b/src/matter/common/MessageExchange.ts
@@ -193,7 +193,7 @@ export class MessageExchange<ContextT> {
         let ackPromise: Promise<void> | undefined;
         if (message.payloadHeader.requiresAck) {
             this.sentMessageToAck = message;
-            this.retransmissionTimer = Time.getTimer(this.getResubmissionBackOffTime(0), () => this.retransmitMessage(message, 1));
+            this.retransmissionTimer = Time.getTimer(this.getResubmissionBackOffTime(0), () => this.retransmitMessage(message, 0));
             const { promise, resolver, rejecter } = await getPromiseResolver<void>();
             ackPromise = promise;
             this.sentMessageAckSuccess = resolver;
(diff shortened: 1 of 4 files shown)


src/matter/common/MessageExchange.ts
```diff
@@ -124,8 +146,11 @@ export class MessageExchange<ContextT> {
             } else {
                 // The other side has received our previous message
                 this.sentMessageAckSuccess?.();
-                this.sentMessageToAck = undefined;
                 this.retransmissionTimer?.stop();
+                this.retransmissionTimer = undefined;
```
Apollon77, line 150: adjust to a "retransmissionRunning" variable

src/matter/common/MessageExchange.ts
```diff
@@ -211,13 +265,22 @@ export class MessageExchange<ContextT> {
             this.send(MessageType.StandaloneAck, new ByteArray(0))
                 .catch(error => logger.error("An error happened when closing the exchange", error));
         }
-        Time.getTimer(this.activeRetransmissionTimeoutMs * (this.retransmissionRetries + 1), () => this.closeInternal())
```
Apollon77, line 214: We need to weait still maximum for Standalone acks and such ... so find maximum

src/matter/common/MessageExchange.ts
```diff
@@ -211,13 +265,22 @@ export class MessageExchange<ContextT> {
...
-            .start();
+        if (this.retransmissionTimer) {
+            this.closeAfterRetransmissionCompleted = true;
+            logger.debug("Close message exchange delayed to after retransmission completed");
+        }
     }
 
-    private closeInternal() {
+    private closeInternal(error?: Error) {
```
Apollon77, line 274: move error and rejection to retransmission

src/matter/common/MessageExchange.ts
Apollon77: or not need it alt all because we wait maximum time

src/matter/session/SecureSession.ts
```diff
@@ -19,6 +25,8 @@ const SESSION_RESUMPTION_KEYS_INFO = ByteArray.fromString("SessionResumptionKeys
 export class SecureSession<T> implements Session<T> {
     private nextSubscriptionId = 0;
     private readonly subscriptions = new Array<SubscriptionHandler>();
+    private timestamp = Date.now();
+    private activeTimestamp = Date.now();
```
Apollon77, line 29: no Date.now -> Timer.now

mfucci (approved):

src/matter/session/SecureSession.ts
```diff
@@ -47,13 +55,24 @@ export class SecureSession<T> implements Session<T> {
         return true;
     }
 
+    touchSessionTimestamps(messageSent: boolean) {
+        this.timestamp = Date.now();
```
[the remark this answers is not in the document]
Apollon77: also replaced by Time.nowMs()

Apollon77:
@mfucci Adjusted as discussed ... and yes there was something wird with the activeTimestamp ... ;-) you were right

PR #241, opened 2023-02-11
Sync Interaction Messages and types with specs

Apollon77:
This PR syncs the Intercartion messages with the specs and adds references

Because of the fact that some types/names are already used in many code parts I decided to add Todos for these refactrings and renamings and would do this once we have a clean code base regaring open PRs and ongoing work.

### DIFF
--- a/src/matter/interaction/InteractionClient.ts
+++ b/src/matter/interaction/InteractionClient.ts
@@ -70,7 +70,7 @@ export class InteractionClient {
     }
 
     async getAllAttributes(): Promise<{}> {
-        return this.withMessenger<{ endpointId: number, clusterId: number, attributeId: number, version: number, value: TlvStream }[]>(async messenger => {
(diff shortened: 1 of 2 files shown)


src/matter/interaction/InteractionClient.ts
```diff
@@ -70,7 +70,7 @@ export class InteractionClient {
     }
 
     async getAllAttributes(): Promise<{}> {
-        return this.withMessenger<{ endpointId: number, clusterId: number, attributeId: number, version: number, value: TlvStream }[]>(async messenger => {
+        return this.withMessenger<{ endpointId: number | undefined, clusterId: number | undefined, attributeId: number | undefined, version: number, value: TlvStream }[]>(async messenger => {
```
mfucci, line 73: I would prefer that we avoid undefined here, and keep todo in field definitions

mfucci (approved):

PR #242, opened 2023-02-11
Do not forget message to ack if we send out "just" a Standalone ack

Apollon77:
When we send out a Standalone Ack because of a Duplicate (or close) we should not "forget" the message which is still in transit and waits to be acked officially, so not clear it.

### DIFF
--- a/src/matter/common/MessageExchange.ts
+++ b/src/matter/common/MessageExchange.ts
@@ -99,11 +99,7 @@ export class MessageExchange<ContextT> {
         if (messageId === this.receivedMessageToAck?.packetHeader.messageId) {
             // Received a message retransmission but the reply is not ready yet, ignoring
             if (requiresAck) {
-                try {
-                    await this.send(MessageType.StandaloneAck, new ByteArray(0));
-                } catch (error) {
-                    logger.error("Failed to send ack", error);
-                }
+                await this.send(MessageType.StandaloneAck, new ByteArray(0));
             }
             return;
         }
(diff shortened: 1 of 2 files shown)


Apollon77:
Ok, was produce weird side effects ,... need to dig deeper

Apollon77:
I open again ... the issues came because Alexa did strange things and tried to commission the device from two echo devices in parallel but then stopped responding on one of em, so in fact errors I saw in log were correct. i will test further but it still seems to be right

src/matter/common/MessageExchange.ts
```diff
@@ -99,7 +99,11 @@ export class MessageExchange<ContextT> {
         if (messageId === this.receivedMessageToAck?.packetHeader.messageId) {
             // Received a message retransmission but the reply is not ready yet, ignoring
             if (requiresAck) {
-                await this.send(MessageType.StandaloneAck, new ByteArray(0));
+                try {
+                    await this.send(MessageType.StandaloneAck, new ByteArray(0));
+                } catch (error) {
```
mfucci, line 104: Can you remove this catch here? Should be catch at top level

mfucci (approved):

src/matter/common/MessageExchange.ts
Apollon77: done

Apollon77: (and checked, yes should be catched elsewhere)

PR #248, opened 2023-02-12
Handle empty Subscription DataReports correctly

Apollon77:
This PR adds the correct handling for "empty" Datareports as result for subscriptions, because according to specs they should set suppressResponse to true. So we also give this info to the messenger to allow a rejection if the other party answers with some ther message (e.g. a "invalid subscription id" message that we need to use to cancel the subscription.


src/matter/common/MessageExchange.ts
```diff
@@ -15,9 +15,21 @@ import { Logger } from "../../log/Logger";
 import { NodeId } from "./NodeId";
 import { ByteArray } from "@project-chip/matter.js";
 import { SecureChannelProtocol } from "../session/secure/SecureChannelProtocol";
+import { MatterError } from "../../error/MatterError";
 
 const logger = Logger.get("MessageExchange");
 
+export class UnexpectedMessageResponseError extends MatterError {
```
mfucci, line 22: UnexpectedMessageError ?
Not clear why there is "Response" here...

src/matter/common/MessageExchange.ts
```diff
@@ -15,9 +15,21 @@ import { Logger } from "../../log/Logger";
...
+    public constructor(
+        public readonly message: string,
+        public readonly data: Message,
+    ) {
+        super();
```
mfucci, line 27: super(`(${MessageCodec.messageToString(data)}) ${message}`);
message is already a parameter of the constructor.

src/matter/common/MessageExchange.ts
```diff
@@ -146,12 +159,25 @@ export class MessageExchange<ContextT> {
...
                 }
             } else {
-                // The other side has received our previous message
-                this.sentMessageAckSuccess?.();
-                this.retransmissionTimer?.stop();
-                this.sentMessageAckSuccess = undefined;
-                this.sentMessageAckFailure = undefined;
-                this.sentMessageToAck = undefined;
+                // If we only expect an Ack without data but got data, reject with this
+                if (this.expectAckOnly && !SecureChannelProtocol.isStandaloneAck(protocolId, messageType)) {
```
mfucci, line 163: This leaks higher layer concerns into the lower layer and it will be really good to avoid this.
You should walk me through what is the intent since this change makes the code very hard to follow.
The MessageExchange should only ensure valid communication at the exchange level, if the flow is not valid for a specific message flow, this should be handled by the higher level controlling that message flow.

mfucci (changes requested):

src/matter/common/MessageExchange.ts
Apollon77: Yes I know but I did not found a better way to do that.
The topic is relevant for cases where a response is normally supressed - then the normal answer is an Standalone Ack. Our current logic will not forward this Standalone Ack to the queue, so there is not way for the "upper layers" to see that, which is ok because the promise gets resolved.

But now the special case comes: if an error occurs the answer is NOT a standalone ack but a StatusResponse message. So the current logic will resolve the promise (as before) and will send the message to the Queue - but the upper layer has no chance to know that there was an "unexpected StatusResponse" as response. SO yes we could so a "getNextMessage" but we need a special logic because we should not wait for any new we just need to check if there is something "already received" (and hope it is ours - no idea if there could be edge cases).

I tried several ideas and all ended in meehh code and this was the best one.

src/matter/common/MessageExchange.ts
```diff
@@ -201,6 +225,9 @@ export class MessageExchange<ContextT> {
             ackPromise = promise;
```
mfucci, line 225: Suggestion: let's have ackPromise return the message containing the message so we can crash if expectAckOnly is set and this message is not a standalone ack.

mfucci (changes requested):

Apollon77:
@mfucci Adjusted as talked

src/matter/common/MessageExchange.ts
```diff
@@ -15,9 +15,19 @@ import { Logger } from "../../log/Logger";
 import { NodeId } from "./NodeId";
 import { ByteArray } from "@project-chip/matter.js";
 import { SecureChannelProtocol } from "../session/secure/SecureChannelProtocol";
+import { MatterError } from "../../error/MatterError";
 
 const logger = Logger.get("MessageExchange");
 
+export class UnexpectedMessageError extends MatterError {
+    public constructor(
+        public readonly message: string,
```
mfucci, line 24: This one doesn't need to be "public readonly" becauae the base class already define it.

src/matter/common/MessageExchange.ts
```diff
@@ -15,9 +15,19 @@ import { Logger } from "../../log/Logger";
...
+        public readonly data: Message,
```
mfucci, line 25: nit: receivedMessage

src/matter/common/MessageExchange.ts
```diff
@@ -206,10 +216,27 @@ export class MessageExchange<ContextT> {
 
         if (ackPromise !== undefined) {
             this.retransmissionTimer?.start();
-            await ackPromise;
-            this.retransmissionTimer?.stop();
-            this.sentMessageAckSuccess = undefined;
-            this.sentMessageAckFailure = undefined;
+            try {
```
mfucci, line 219: const { payloadHeader: {protocolId, messageType} } = await ackPromise;
if (expectAckOnly && !SecureChannelProtocol.isStandaloneAck(protocolId, messageType)) throw new UnexpectedMessageError([...])

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -190,7 +190,20 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
             }
         }
 
-        await this.exchange.send(MessageType.ReportData, TlvDataReport.encode(dataReport));
+        try {
```
mfucci, line 193: if (dataReport.suppressResponse) {
  tryCatch(() => await this.exchange.send(MessageType.ReportData, true),
     UnexpectedMessageError, error => this.throwIfError(messageType, 
message.payload);
    // add comment
} else {
   await this.exchange.send(MessageType.ReportData);
   await this.waitForSuccess();
}

Apollon77:
Ready for re-review

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -190,7 +191,16 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
             }
         }
 
-        await this.exchange.send(MessageType.ReportData, TlvDataReport.encode(dataReport));
+        if (dataReport.suppressResponse) {
+            // We do not expect a response other than a Standalone Ack, so if we receive anything else, we throw an error
+            await tryCatchAsync(async () => await this.exchange.send(MessageType.ReportData, TlvDataReport.encode(dataReport), true), UnexpectedMessageError, error => {
+                const { payloadHeader: { messageType }, payload } = (error as UnexpectedMessageError).receivedMessage;
```
mfucci, line 197: tryCatchAsync needs to correctly bind the expected type with the argument of the handler, so you should not have to class cast here.

Modify this:
type ErrorHandler<E extends Error, T> = (error: E) => T;

and

export async function tryCatchAsync<E extends Error, T>(codeBlock: () => Promise<T>, errorType: ClassExtends<E>, fallbackValueOrFunction: ErrorHandler<E, T> | T): Promise<T> {

mfucci (changes requested):

src/matter/interaction/InteractionMessenger.ts
Apollon77: Cool, that solved a lot ... included now ...

Apollon77:
Ready for re-review

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -194,7 +194,7 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
         if (dataReport.suppressResponse) {
             // We do not expect a response other than a Standalone Ack, so if we receive anything else, we throw an error
             await tryCatchAsync(async () => await this.exchange.send(MessageType.ReportData, TlvDataReport.encode(dataReport), true), UnexpectedMessageError, error => {
-                const { payloadHeader: { messageType }, payload } = (error as UnexpectedMessageError).receivedMessage;
+                const { payloadHeader: { messageType }, payload } = error.receivedMessage;
```
mfucci, line 197: nit, you can also de-structure receivedMessage
const { receivedMessage: { payloadHeader: { messageType }, payload } }= error;

or even better, you can de-structure in the method argument:
( { receivedMessage: { payloadHeader: { messageType }, payload } }) => this.throwIfError(messageType, payload)

or even better, you can modify throwIfError to take the message as an argument:
({ receivedMessage }) => this.throwIfError(receivedMessage)
(and probably throwIfError should be renamed to throwIfErrorStatusMessage)

mfucci (approved):

src/matter/interaction/InteractionMessenger.ts
Apollon77: Changed to last proposal and also adjusted in second throwIfError case (SecureChannelmessenger) accordingly

PR #249, opened 2023-02-12
Catch StatusResponse exceptions on invoke and send this status

Apollon77:
This PR uses the option that a invoked command can throw a "StatusResponseError" with a defined code and when this code is used and send out.

I decided to not use the tryCatch approach here because the error handler needs to be async and needs to await the Status send action (the alternative would be strange code in my eyes).

### DIFF
--- a/src/matter/interaction/InteractionMessenger.ts
+++ b/src/matter/interaction/InteractionMessenger.ts
@@ -157,7 +157,7 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                 logger.info(`Sending status response ${error.code} for interaction error: ${error}`);
                 await this.sendStatus(error.code);
             } else {
-                logger.error(error.stack ?? error);
+                logger.error(error);
                 await this.sendStatus(StatusCode.Failure);
             }
         } finally {
--- a/src/matter/interaction/InteractionMessenger.ts
+++ b/src/matter/interaction/InteractionMessenger.ts
@@ -153,14 +153,13 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                 }
             }
         } catch (error: any) {
-            let statusCode = StatusCode.Failure;
             if (error instanceof StatusResponseError) {
-                statusCode = error.code;
+                logger.info(`Sending status response ${error.code} for interaction error: ${error}`);
+                await this.sendStatus(error.code);
             } else {
                 logger.error(error.stack ?? error);
+                await this.sendStatus(StatusCode.Failure);
             }
-            logger.info(`Sending status response ${statusCode} for interaction error: ${error}`);
-            await this.sendStatus(statusCode);
         } finally {
             this.exchange.close();
         }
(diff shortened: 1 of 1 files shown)


src/matter/interaction/InteractionMessenger.ts
```diff
@@ -139,8 +139,18 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                         break;
                     case MessageType.InvokeCommandRequest:
                         const invokeRequest = TlvInvokeRequest.decode(message.payload);
-                        const invokeResponse = await handleInvokeRequest(invokeRequest, message);
-                        await this.exchange.send(MessageType.InvokeCommandResponse, TlvInvokeResponse.encode(invokeResponse));
+                        try {
+                            const invokeResponse = await handleInvokeRequest(invokeRequest, message);
+                            await this.exchange.send(MessageType.InvokeCommandResponse, TlvInvokeResponse.encode(invokeResponse));
+                        } catch (error) {
+                            if (error instanceof StatusResponseError) {
```
mfucci, line 146: I think this should be in the already existing catch statement below, nothing specific to InvokeCommandRequest

mfucci (changes requested):

src/matter/interaction/InteractionMessenger.ts
Apollon77: Good Idea, adjusted

Apollon77:
@mfucci adjusted

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -153,8 +153,14 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                 }
             }
         } catch (error: any) {
-            logger.error(error.stack ?? error);
-            await this.sendStatus(StatusCode.Failure);
+            let statusCode = StatusCode.Failure;
```
mfucci, line 156: Quite hard to read. Maybe if (error instanceof StatusResponseError) and inline both code paths?

src/matter/interaction/InteractionMessenger.ts
```diff
@@ -153,8 +153,13 @@ export class InteractionServerMessenger extends InteractionMessenger<MatterDevic
                 }
             }
         } catch (error: any) {
-            logger.error(error.stack ?? error);
-            await this.sendStatus(StatusCode.Failure);
+            if (error instanceof StatusResponseError) {
+                logger.info(`Sending status response ${error.code} for interaction error: ${error}`);
+                await this.sendStatus(error.code);
+            } else {
+                logger.error(error.stack ?? error);
```
mfucci, line 160: Let's avoid using error.stack :
1 _ Per this [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack), this is a non-standard property
2 _ There is no reason to use it only here versus every other logger.error(error) calls. So this behavior should be implemented inside the Logger to be consistent across the board
```suggestion
                logger.error(error);
```

Apollon77: Sure (BTW: That was from you before, I did not change that at all. :-)) But yes in the meantime we also have handling in the logger for that

PR #250, opened 2023-02-12
Fix several typos and var usage

Apollon77:


src/matter/cluster/server/OperationalCredentialsServer.ts
```diff
@@ -21,7 +21,7 @@ import { tryCatch } from "../../../error/TryCatchHandler";
 import { FabricNotFoundError } from "../../fabric/FabricManager";
 import { Logger } from "../../../log/Logger";
 
-const logger = Logger.get("OperationalCredentialsServer");
+const logger = Logger.get("OperationalCredentialClusterServer");
```
mfucci, line 24: This doesn't match the file name. If the file name is incorrect, it should also be updated.

mfucci (approved):

src/matter/cluster/server/OperationalCredentialsServer.ts
Apollon77: renamed

PR #251, opened 2023-02-12
Optimize Fabric deletion and destroy sessions

Apollon77:
When a Fabric gets deleted the current session associated with the Fabric needs to be destroyed. If it is the last fabric all other remaining sessions also needs to be destroyed


src/matter/MatterDevice.ts
```diff
@@ -91,6 +103,10 @@ export class MatterDevice {
         this.fabricManager.removeFabric(fabricIndex);
     }
 
+    getFabricbyIndex(fabricIndex: FabricIndex) {
```
mfucci, line 106: nit: getFabricByIndex

src/matter/cluster/server/OperationalCredentialsServer.ts
```diff
@@ -119,7 +119,13 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
 
                 // TODO persist fabrics
                 // TODO: depending on cases destroy the secure session and delete all data!
-
+                if (device.getFabrics().length === 0) {
+                    logger.info(`Expire all sessions because deleted fabric ${fabricIndex.index} was the last one`);
```
mfucci, line 123: Why do this?
The code below should bring to the same state since there will be no more Fabrics

src/matter/session/SessionManager.ts
```diff
@@ -59,11 +60,38 @@ export class SessionManager<ContextT> {
...
+            if (!session.isSecure()) return false;
+            const secureSession = session as SecureSession<any>;
+            if (secureSession.getFabric()?.fabricIndex.index === fabricIndex.index) {
+                session.destroy();
+                this.sessions.delete(session.getId());
+            }
+        });
+    }
+
+    destroyAllSessions() {
```
mfucci, line 82: This should never be called: UNICAST_UNSECURE_SESSION_ID pointing to unsecureSession is added in the constructor and this code will delete it.

What is the purpose of this method? I am not sure I understand when all the sessions need to be destroyed

src/matter/cluster/server/OperationalCredentialsServer.ts
```diff
@@ -119,7 +119,13 @@ export const OperationalCredentialsClusterHandler: (conf: OperationalCredentials
```
mfucci, line 119: Removing session associated with a fabric when a fabric is deleted should be done by the FabricManager, otherwise there is too much risk to have the system in an incoherent state.

src/matter/session/SessionManager.ts
```diff
@@ -59,11 +60,38 @@ export class SessionManager<ContextT> {
...
 
+    destroySession(sessionId: number) {
+        const session = this.sessions.get(sessionId);
+        if (session) {
+            session.destroy();
+            this.sessions.delete(sessionId);
+        }
+    }
+
+    destroySessionForFabricIndex(fabricIndex: FabricIndex) {
```
mfucci, line 71: This should done be the FabricManager, add a getSessionsForFabric here and the FabricManager will be in charge of calling destroySession. Each manager should take of his own mess...

src/matter/session/UnsecureSession.ts
```diff
@@ -77,4 +77,8 @@ export class UnsecureSession<T> implements Session<T> {
     getPeerNodeId() {
         return undefined;
     }
+
+    destroy() {
+        // Nothing to do
```
mfucci, line 82: throw an Error here, this session should never be destroyed

mfucci (changes requested):

src/matter/session/SessionManager.ts
Apollon77: you are right ... I remive it again

src/matter/cluster/server/OperationalCredentialsServer.ts
Apollon77: right

Apollon77: PS: The specs are phrasing the case a bit different but for now you are tight... it is the same

src/matter/cluster/server/OperationalCredentialsServer.ts
Apollon77: I moved it to "MatterDevcie" now because FabricManager do not have acces to the session controller but the device is taling care here too ... so should work? 

ALternatively Ineed to pass the sessionmanager with removeFabric justto call from there ...

Apollon77:
@mfucci COmments addressed ... one topic basically open - see above, but I did a change to where it also could make sense in my eyes. Ready for re-review

mfucci (commented):
Let's make sure the system cannot fall in an incorrect state

Apollon77:
@mfucci Adjusted as discussed

src/matter/fabric/Fabric.ts
```diff
@@ -55,6 +60,21 @@ export class Fabric {
...
+    }
+
+    setRemoveCallback(callback: () => void) {
+        this.removeCallback = callback;
+    }
+
+    remove() {
+        for (const session of this.sessions) {
+            session.destroy();
+        }
```
mfucci, line 75:
```suggestion
        this.sessions.forEach(session => session.destroy());
```

src/matter/session/SecureSession.ts
```diff
@@ -32,12 +32,14 @@ export class SecureSession<T> implements Session<T> {
...
+    static async create<T>(context: T, id: number, fabric: Fabric | undefined, peerNodeId: NodeId, peerSessionId: number, sharedSecret: ByteArray, salt: ByteArray, isInitiator: boolean, isResumption: boolean, closeCallback: () => void, idleRetransTimeoutMs?: number, activeRetransTimeoutMs?: number) {
         const keys = await Crypto.hkdf(sharedSecret, salt, isResumption ? SESSION_RESUMPTION_KEYS_INFO : SESSION_KEYS_INFO, 16 * 3);
         const decryptKey = isInitiator ? keys.slice(16, 32) : keys.slice(0, 16);
         const encryptKey = isInitiator ? keys.slice(0, 16) : keys.slice(16, 32);
         const attestationKey = keys.slice(32, 48);
-        return new SecureSession(context, id, fabric, peerNodeId, peerSessionId, sharedSecret, decryptKey, encryptKey, attestationKey, idleRetransTimeoutMs, activeRetransTimeoutMs);
+        const session = new SecureSession(context, id, fabric, peerNodeId, peerSessionId, sharedSecret, decryptKey, encryptKey, attestationKey, closeCallback, idleRetransTimeoutMs, activeRetransTimeoutMs);
```
mfucci, line 40: You also need to remove the session on the fabric.

=> better to move
fabric?.addSession(session);
and
fabric?.removeSession(sessionId);

inside the session itself

mfucci (changes requested):

src/matter/session/SecureSession.ts
Apollon77: You are rught, forgot that one. and also another place. Adjusted

mfucci (approved):

PR #254, opened 2023-02-15
Fix for OSX network interface.  set to en0

vves:


JimBuzbee:
This won't work for one of my Macs that is currently hooked up to ethernet via a USB-C docking station. It's using en10.

# route get default | grep interface | awk '{print $2}'

I'd rather see the interface optional passed in via a parameter, e.g. something like:

# node Device.js -interface $(route get default | grep interface | awk '{print $2}')

If "-interface" not used, it would act as it does now

Apollon77:
I also think that this is way to specific. The network interfaces are very dynamic and deopend on many factors. en0 csn be anything.

So @vves what exactly you have as error?

I also curremtly look into that aside and gathered some ideas today https://github.com/mfucci/node-matter/issues/11#issuecomment-1432108984

jdesai61:
I mentioned the same in #11, but will repeat here, there should be an optional command line argument to specify interface and perhaps even the IP.  If the standard specifies IPV6 only, perhaps support for IPV4 can be dropped

Apollon77:
Yes I will also consider this in my coming PR

vves:
Note the Multicast Server error below  I have three OSX boxes across two users of this library that this is 'enough' so that the library works out of the box.  I agree that better command line reference would be useful, but I would think most consumers of this library including my team will be implementing their own 'Device' or 'Bridge' so I take Device.ts as a template more than a final implementation.  Is it worth adding more command line params or just make the demo run out-of-the-box as the case may be.  'en0' is always defined on OSX as ethernet if it exists and wireless if a wired connection is not available.  This method provides a narrow fix for OSX users without changing the behavior for other platforms.

QR Code URL: https://project-chip.github.io/connectedhomeip/qrcode.html?data=MT:Y.K90AFN00KA0648G00
Manual pairing code: 34970112332
2023-02-15 16:47:19.094 INFO UdpMulticastServer utun2: send EMSGSIZE ff02::fb:5353

jdesai61:
It's not an OSX only issue. I am seeing many interfaces on linux VMs as well as RPIs where I am testing node-matter.



Sent from my mobile device
________________________________

vves:
@jdesai61 - thanks for the updated information.  I have a three different linux flavor VMs but had only tested on one and it was fine.  I'll double check with the other two.  I still question if the Device.js is just a template for node-matter consumers to use as a template how far down the rabbit hole of fixing this we should go.  BRB with results from my other VMs.

Apollon77:
I would more tend to solve it "correctly" ...

vves:
Note.  Both my Debian 11.3 and Ubuntu 22.04 VM instances provide IPv4 address before the IPv6 in the ips array and iOS Home App does not find the Matter Virtual Device and start the commissioning process.

Curiously - the MdnsBroadcaster changes IP addresses on Debian Linux after ~10 seconds to - 
`MdnsBroadcaster on IP Addr: 127.0.0.1,::1 & hostname: 0000000000000000.local`

On OSX, the IPv6 address is listed first in the array and Home App is able to pair:   
`MdnsBroadcaster on IP Addr: fe80::1420:195b:ad8c:8262,10.0.10.145 & hostname: F02F4B09EDFA0000.local`

Not sure if this has anything to do with the Home App not being able to locate the Virtual Device on the network or if it's a red-herring.  Just more information.

vves:
@jdesai61 I was able to confirm that using a Belkin hub that en4 becomes this interface on OSX..  I am removing pull request for now.

Apollon77:
I will have a PR soon but after playing around two days with that topic I just add to allow to select the interface to send out the annoucemements ... In fact all "receiving" parts need to listen to anything and in fact the UDP traffic  can come in on any interface in the end. As soon as you limit this you end is issues over and over

PR #256, opened 2023-02-19
Initialize Secure Case Session earlier

Apollon77:
This PR initialize the SecureSession instance a bit earlier because we send the sessionId already out as Sigma2Resule before we initialize the session, WHile playing around I had edge cases where the peer sent in requests with the new ID before we had registered the session id internally in the SessionManager. 

Moving the code solvs this. The worst case is that Sigma2Result fails (which can normally only happen in network issue cases) and we have the session object created which is then not used

### DIFF
--- a/src/matter/session/secure/CaseServer.ts
+++ b/src/matter/session/secure/CaseServer.ts
@@ -45,10 +45,16 @@ export class CaseServer implements ProtocolHandler<MatterDevice> {
         // Try to resume a previous session
         const resumptionId = Crypto.getRandomData(16);
         let resumptionRecord;
+
+        // We try to resume the session
         if (peerResumptionId !== undefined && peerResumeMic !== undefined && (resumptionRecord = server.findResumptionRecordById(peerResumptionId)) !== undefined) {
             const { sharedSecret, fabric, peerNodeId } = resumptionRecord;
             const peerResumeKey = await Crypto.hkdf(sharedSecret, ByteArray.concat(peerRandom, peerResumptionId), KDFSR1_KEY_INFO);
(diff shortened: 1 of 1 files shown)


src/matter/session/secure/CaseServer.ts
```diff
@@ -45,20 +45,23 @@ export class CaseServer implements ProtocolHandler<MatterDevice> {
...
 
+            // All good! Create secure session
+            const secureSessionSalt = ByteArray.concat(peerRandom, peerResumptionId);
+            const secureSession = await server.createSecureSession(sessionId, fabric, peerNodeId, peerSessionId, sharedSecret, secureSessionSalt, false, true, mrpParams?.idleRetransTimeoutMs, mrpParams?.activeRetransTimeoutMs);
+
             // Generate sigma 2 resume
             const resumeSalt = ByteArray.concat(peerRandom, resumptionId);
             const resumeKey = await Crypto.hkdf(sharedSecret, resumeSalt, KDFSR2_KEY_INFO);
             const resumeMic = Crypto.encrypt(resumeKey, new ByteArray(0), RESUME2_MIC_NONCE);
             await messenger.sendSigma2Resume({ resumptionId, resumeMic, sessionId });
```
mfucci, line 63: Add a try / catch to destroy session in case of error.
Check if the same should be done in Pase

Apollon77: done

Apollon77:
Session is now destroyed on error, #251 mkes sure that session is then also removed from sessionmanager

mfucci (approved):

PR #257, opened 2023-02-19
Controller/InteractionClient Process optimizations and refactoring

Apollon77:
This PR collects some things that I could not split better. Ithink it might be easier to review the single Commits because they for themself are simpler.

In fact the PR does:
* Cleanup in ChannelManager (especially centrallizing the Channelkey generation. It also adds "removeChannel" which is needed when InteractionCLient is reconnecting
* Create and throw own typed error for Retransmission Reached case to allow more easy catching
* Make timeout to discover a device configurable - during commissioning we usually wait 5s, for resuming a session (preparation for persistence!) we should allow to wait longer, so I made the timeout configurable
* Add session-Reconnect logic to "withMessanger" for Interactionclient to renew session - If we have a running connection to a device in controller case and the device gets restarted then the current session becomes invalid. Apple/Google/Amazon are opening a new session when this happens so we simply do the same. The "withController" method which is used gets enhances by a method that also can handle a reconnect if the Retransmission-Reached Error is thrown for the send command.
* The interactionClient gets an optional "reconnect" callback for this case to have the logic itself in MatterController where the rest of the connection login is owned. I think thisan elegant way to keep the topics selarated - if you have a better one just tell.
* The InteractionClient now also gets a method to "resume" a session and because we do something like that also when doing CASE pairing I refactored to code that the "connect" method takes care of that. It will all be used later when persistence is there - then the code can know if we have are commissioned or not and depending on this enter commissioning or just do a session resume/reconnect. So this is here preparation


src/matter/mdns/MdnsScanner.ts
```diff
@@ -38,15 +38,15 @@ export class MdnsScanner implements Scanner {
         this.periodicTimer = Time.getPeriodicTimer(60 * 1000 /* 1 mn */, () => this.expire()).start();
     }
 
-    async findDevice({operationalId}: Fabric, nodeId: NodeId): Promise<MatterServer | undefined> {
+    async findDevice({operationalId}: Fabric, nodeId: NodeId, timeOut = 5): Promise<MatterServer | undefined> {
```
mfucci, line 41: timeOut => timeOutSeconds

src/matter/interaction/InteractionClient.ts
```diff
@@ -81,13 +81,14 @@ export class InteractionClient {
 
     constructor(
         private readonly exchangeManager: ExchangeManager<MatterController>,
```
mfucci, line 83: I would recommend to add another layer to handle those reconnect.
messageExchangeProvider: () => Promise<MessageExchange>
and the recoonect logic is implemented in the provider

src/matter/common/ChannelManager.ts
```diff
@@ -11,28 +11,39 @@ import { Session } from "../session/Session";
...
 import { NodeId } from "./NodeId";
 import { ByteArray } from "@project-chip/matter.js";
+import { MatterError } from "../../error/MatterError";
+
+export class NoChannelError extends MatterError {};
 
 export class ChannelManager {
     private readonly channels = new Map<string, MessageChannel<any>>();
 
+    getChannelKey(fabric: Fabric, nodeId: NodeId) {
```
mfucci, line 21: Make it private

Apollon77:
@mfucci Ready for re-review ... but also see comment above

src/matter/interaction/InteractionClient.ts
Apollon77: Ok, @mfucciOk, check nwo ... Ithink this is the "best" place now ... we can chat about "naming" of that one new class.

In fact the "InteractionClientMessanger" is now handling this with the ExchangeProvider and so we do it on the "exchange.send" level, but just in case that we have the case that WE initialize the session. For InteractionServer (and also InteractionCLient for Subscriptionhandler" the exchanges are not build by us but initiated by an incoming connection, so here we can not do the retry.

WDYT?

src/matter/interaction/InteractionClient.ts
```diff
@@ -76,14 +76,35 @@ export class SubscriptionClient implements ProtocolHandler<MatterController> {
...
 export class InteractionClient {
     private readonly subscriptionListeners = new Map<number, (dataReport: DataReport) => void>();
 
     constructor(
-        private readonly exchangeManager: ExchangeManager<MatterController>,
-        private readonly channel: MessageChannel<MatterController>,
+        private readonly exchangeProvider: ExchangeProvider,
     ) {
-        this.exchangeManager.addProtocolHandler(new SubscriptionClient(this.subscriptionListeners));
+        this.exchangeProvider.addProtocolHandlerToExchangeManager(new SubscriptionClient(this.subscriptionListeners));
```
mfucci, line 107: This doesn't right... We might add the SubscriptionClient multiple time on the exchangeManager.

The ExchangeManager should be a singleton for the Device, all protocol handlers should only be added during the initialization and the list not modified afterwards.

Try to restructure the code so this is added only once.
Or we should discuss this during a review session

mfucci (changes requested):

src/matter/interaction/InteractionClient.ts
Apollon77: I do not see why we wpuld add it multiple times - and if then the issue existed before already. If you see such cases then we should address them in a separate PR.

And yes many of this will be reworked again when we do the new Hight level API and such ... lets cleanup then please

Apollon77: @mfucci Ok I checked it again.

First of all it is Controller and not Device

We have currently two places where we instanciate an InteractionClient and in the past the exchangeManager was handed in and SubscriptionClient registered. So yes it is very good possible that the SubscriptionHandler was and is registered potentiella multiple times.

Because the issue existed already before I would add a Todo and address this after the move

Apollon77: PS: And when we do high level API we need to adjust that also in any case (again)

Apollon77:
Reintroduced in matter.js

PR #262, opened 2023-02-24
Generate PAI, DAC and CertificateDeclaration

Apollon77:
The PR generates the PAI, DAC and also CertificationDeclaration based on the set vendorId and productId.

The only "static" values now are the PAA (Authority) and the Private key used for the CD

This sitested with Apple, Google, Alexa (that all basically ignored issues with the certs/CD) and Chip-Tool (which cares a lot :-) )


Apollon77:
@turon If you like you are also welcome to review

mfucci (approved):

PR #266, opened 2023-03-08
Initial implementation of the Power Source Configuration Cluster

JimBuzbee:
Tiny cluster for describing the power source configuration and capabilities

### DIFF
--- a/src/matter/cluster/PowerSourceConfigurationCluster.ts
+++ b/src/matter/cluster/PowerSourceConfigurationCluster.ts
@@ -27,7 +27,7 @@ export const PowerSourceConfigurationCluster = Cluster({
     /** @see {@link MatterCoreSpecificationV1_0} § 11.6.5.1 */
     attributes: {
         /** The set of all power sources capable of participating in the power system of this Node. */
-        sources: Attribute(0x0, TlvArray(TlvEndpointNumber), { default: [] }), // max 6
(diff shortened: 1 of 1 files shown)


src/matter/cluster/PowerSourceConfigurationCluster.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export const PowerSourceConfigurationCluster = Cluster({
+    id: 0x2e,
+    name: "Power Source Configuration",
+    revision: 1,
+
+    /** @see {@link MatterCoreSpecificationV1_0} § 11.6.5.1 */
+    attributes: {
+        /** The set of all power sources capable of participating in the power system of this Node. */
+        sources: Attribute(0x0, TlvArray(TlvEndpointNumber), { default: [] }), // max 6
```
Apollon77, line 30: we should have maxLength: 6 as property to define maxLength

Apollon77 (changes requested):

src/matter/cluster/PowerSourceConfigurationCluster.ts
JimBuzbee: Thanks - Couldn't figure out the syntax earlier, but got it this time and re-pushed

Apollon77 (approved):

PR #268, opened 2023-03-10
Recheck and sync all clusters again against Chip XML/Specs

Apollon77:
I did a full check of all clusters in as of now against the Chip XMLs and specs ... Here are the results.

I also streamlined all names tp be "without spaces" so that it easiuer readable in logging

### DIFF
--- a/src/matter/cluster/Cluster.ts
+++ b/src/matter/cluster/Cluster.ts
@@ -98,7 +98,7 @@ export interface Cluster<F extends BitSchema, A extends Attributes, C extends Co
     id: number,
     name: string,
     revision: number,
-    features: BitSchema,
+    features: F,
     attributes: A,
     commands: C,
     events: E,
--- a/src/matter/cluster/FlowMeasurementCluster.ts
+++ b/src/matter/cluster/FlowMeasurementCluster.ts
@@ -23,11 +23,11 @@ export const FlowMeasurementCluster = Cluster({
        measuredValue: Attribute(0, TlvNullable(TlvUInt16), { default: null }),
 
        /** Indicates the minimum value of MeasuredValue that can be measured.  */
-       minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({ min: 0 }))),
+       minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16)),
 
        /** Indicates the maximum value of MeasuredValue that can be measured.  */
        maxMeasuredValue: Attribute(2, TlvNullable(TlvUInt16.bound({ max: 65534 }))),
 
-       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ }), { default: 0 }),
+       tolerance: OptionalAttribute(3, TlvUInt16.bound({ max: 2048 /* 0x0800 */ }), { default: 0 }),
     },
 });
--- a/src/matter/cluster/FlowMeasurementCluster.ts
+++ b/src/matter/cluster/FlowMeasurementCluster.ts
@@ -21,13 +21,13 @@ export const FlowMeasurementCluster = Cluster({
     attributes: {
        /** MeasuredValue represents the flow in m3 /h as follows: MeasuredValue = 10 x Flow  */
        measuredValue: Attribute(0, TlvNullable(TlvUInt16), { default: null }),
-       
+
        /** Indicates the minimum value of MeasuredValue that can be measured.  */
        minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({ min: 0 }))),
-       
+
        /** Indicates the maximum value of MeasuredValue that can be measured.  */
        maxMeasuredValue: Attribute(2, TlvNullable(TlvUInt16.bound({ max: 65534 }))),
 
-       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 }), { default: 0 }),
+       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ }), { default: 0 }),
     },
 });
--- a/src/matter/cluster/IlluminanceMeasurementCluster.ts
+++ b/src/matter/cluster/IlluminanceMeasurementCluster.ts
@@ -21,14 +21,14 @@ export const IlluminanceMeasurementCluster = Cluster({
     attributes: {
        /** Represents the illuminance in Lux (symbol lx): MeasuredValue = 10,000 x log10(illuminance) + 1 */
        measuredValue: Attribute(0, TlvNullable(TlvUInt16), { default: 0 }),
-       
+
        /** Indicates the minimum value of MeasuredValue that can be measured. */
        minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({ min: 1 }))),
-       
+
        /** Indicates the maximum value of MeasuredValue that can be measured. */
        maxMeasuredValue: Attribute(2, TlvNullable(TlvUInt16.bound({ max: 65534 }))),
-       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 })),
-       
+       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ })),
+
        /** Specifies the electronic type of the light sensor. */
        lightSensorType: OptionalAttribute(4, TlvNullable(TlvUInt8), { default: null }), // only values null, 0, 1 and 0x40 to 0xfe are allowed
     },
--- a/src/matter/cluster/IlluminanceMeasurementCluster.ts
+++ b/src/matter/cluster/IlluminanceMeasurementCluster.ts
@@ -27,7 +27,7 @@ export const IlluminanceMeasurementCluster = Cluster({
 
        /** Indicates the maximum value of MeasuredValue that can be measured. */
        maxMeasuredValue: Attribute(2, TlvNullable(TlvUInt16.bound({ max: 65534 }))),
-       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ })),
+       tolerance: OptionalAttribute(3, TlvUInt16.bound({ max: 2048 /* 0x0800 */ })),
 
        /** Specifies the electronic type of the light sensor. */
        lightSensorType: OptionalAttribute(4, TlvNullable(TlvUInt8), { default: null }), // only values null, 0, 1 and 0x40 to 0xfe are allowed
--- a/src/matter/cluster/LevelControlCluster.ts
+++ b/src/matter/cluster/LevelControlCluster.ts
@@ -154,7 +154,7 @@ export const LevelControlCluster = Cluster({
         startUpCurrentLevel: OptionalWritableAttribute(0x4000, TlvNullable(TlvUInt8), { persistent: true }),
     },
     commands: commonCommands,
- });
+});
 
  /**
  * Attributes and commands for Pulse Width Modulation (Provisional)
@@ -184,4 +184,4 @@ export const LevelControlCluster = Cluster({
         /** Changes current frequency to the requested frequency, or to the closest frequency. */
         moveToClosestFrequency: Command(0x8, MoveToClosestFrequencyCommandRequest, 0x8, TlvNoResponse),
     },
- });
+});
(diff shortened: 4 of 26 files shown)


src/matter/cluster/LevelControlCluster.ts
```diff
@@ -55,23 +55,23 @@ const MoveToClosestFrequencyCommandRequest = TlvObject({
...
     /** Dependency with the On/Off cluster */
-    onOffClusterDependency: BitFlag(0), // default: true
+    onOff: BitFlag(0), // default: true
 
     /** Behavior that supports lighting applications. */
     lighting: BitFlag(1),
 
     /** Supports frequency attributes and behavior. */
-    supportFrequency: BitFlag(2)
+    frequency: BitFlag(2)
```
JimBuzbee, line 64: I originally had this as "frequency" and also "onOff" as in the spec, but changed it based on comment from mfucci who wanted these to be more descriptive.

JimBuzbee (approved):
Nice work to make everything consistent, and you found a few bugs along the way!

src/matter/cluster/LevelControlCluster.ts
Apollon77: We should be close to specs and XML (this is also something we agreed on. With the comment line over it it should be clear also in IDE when using it

JimBuzbee: I'd tend to agree that unless there is a technical reason not to, we should stick to the spec even when it could have been written better.

src/matter/cluster/Cluster.ts
```diff
@@ -98,7 +98,7 @@ export interface Cluster<F extends BitSchema, A extends Attributes, C extends Co
     id: number,
     name: string,
     revision: number,
-    features: BitSchema,
+    features: F,
```
Apollon77, line 101: @mfucci I assume it should be that way, right?

src/matter/cluster/LevelControlCluster.ts
```diff
@@ -55,23 +55,23 @@ const MoveToClosestFrequencyCommandRequest = TlvObject({
...
 };
 
 /** @see {@link MatterApplicationClusterSpecificationV1_0} § 1.6.6.2.1 */
-const enum MoveMode {
+export const enum MoveMode {
```
Apollon77, line 68: Needed if we want to use it to send or code with ... so experted it now because needed in my local version. I could revert here but I assume that it do not make that big of a difference

src/matter/cluster/Cluster.ts
mfucci: Correct!

src/matter/cluster/LevelControlCluster.ts
mfucci: OK.... I asked to rename those since the spec names were pretty bad, but fine...

src/matter/cluster/OnOffCluster.ts
```diff
@@ -97,23 +119,23 @@ export const OnOffCluster = Cluster({
 
         // The following attributes are only needed for "Level Control for Lighting" support
 
-        /** Used to remember if a state is already storednin an old scene to not store one again when sending another Off command */
-        //globalSceneControl: OptionalAttribute(0x4000,  TlvBoolean, { default: true }),
+        /** Used to remember if a state is already stored in an old scene to not store one again when sending another Off command */
+        globalSceneControl: OptionalAttribute(0x4000,  TlvBoolean, { default: true }),
```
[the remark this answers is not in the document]
mfucci: Agreed with Ingo: globalSceneControl is not optional.
It should not be present when features.lightingLevelControl = false
and it is mandatory when features.lightingLevelControl = true

Let's keep those as commented out for now and we need to figure out a way to have the cluster definition changing based on a feature flag value.

src/matter/cluster/PressureMeasurementCluster.ts
```diff
@@ -45,7 +45,7 @@ export const PressureMeasurementCluster = Cluster({
        maxScaledValue: OptionalAttribute(0x12, TlvNullable(TlvInt16), { default: 0 }),
 
        /** Indicates the magnitude of the possible error that is associated with ScaledValue */
-       scaledTolerance: OptionalAttribute(0x13, TlvUInt16.bound({ min: 0, max: 2048 }), { default: 0 }),
+       scaledTolerance: OptionalAttribute(0x13, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ }), { default: 0 }),
 
        /** Indicates the base 10 exponent used to obtain ScaledValue */
        scale: OptionalAttribute(0x14, TlvUInt8.bound({ min: -127, max: 127 }), { default: 0 }),
```
mfucci, line 51: This is incorrect, TlvUInt8 is an unsigned int so cannot be below 0. Should this be a signed int?

src/matter/cluster/TemperatureMeasurementCluster.ts
```diff
@@ -20,7 +20,7 @@ export const TemperatureMeasurementCluster = Cluster({
     /** @see {@link MatterApplicationClusterSpecificationV1_0} § 2.3.4 */
     attributes: {
        /** MeasuredValue = 100 x temperature [°C] **/
-       measuredValue: Attribute(0, TlvNullable(TlvInt16)),
+       measuredValue: Attribute(0, TlvNullable(TlvInt16.bound({ min: -27315, max: 32767 }))),
```
mfucci, line 23: 32767 is already the maximum for an TlvInt16 so no need to bound it again (but min is lower)

src/matter/cluster/WaterContentMeasurementCluster.ts
```diff
@@ -11,19 +11,19 @@ import {BitFlag, MatterApplicationClusterSpecificationV1_0, TlvBitmap, TlvInt16,
 /** @see {@link MatterApplicationClusterSpecificationV1_0} § 2.6.4 */
 const  attributes = {
     /** Represents the water content in % as follows: MeasuredValue = 100 x water content */
-    measuredValue: Attribute(0, TlvNullable(TlvUInt16.bound({min: 0, max: 10000}))),
+    measuredValue: Attribute(0, TlvNullable(TlvUInt16.bound({ min: 0, max: 10000 }))),
```
mfucci, line 14: min is already 0 since this is an unsigned int

src/matter/cluster/WaterContentMeasurementCluster.ts
```diff
@@ -11,19 +11,19 @@ import {BitFlag, MatterApplicationClusterSpecificationV1_0, TlvBitmap, TlvInt16,
...
 
     /** Indicates the minimum value of MeasuredValue that can be measured. */
-    minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({min: 0}))),
+    minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({ min: 0 }))),
```
mfucci, line 17: min is already 0 since this is an unsigned int

src/matter/cluster/WaterContentMeasurementCluster.ts
```diff
@@ -11,19 +11,19 @@ import {BitFlag, MatterApplicationClusterSpecificationV1_0, TlvBitmap, TlvInt16,
...
 
-    /** 
-     * The magnitude of the possible error that is associated with MeasuredValue 
-     * attribute using the same unit 
+    /**
+     * The magnitude of the possible error that is associated with MeasuredValue
+     * attribute using the same unit
      */
-    tolerance: OptionalAttribute(3, TlvUInt16.bound({min: 0, max: 2048})),
+    tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ })),
```
mfucci, line 26: min is already 0 since this is an unsigned int

src/matter/cluster/PressureMeasurementCluster.ts
```diff
@@ -18,22 +18,22 @@ export const PressureMeasurementCluster = Cluster({
...
        /** Indicates the minimum value of MeasuredValue that can be measured. */
        minMeasuredValue: Attribute(0x1, TlvNullable(TlvInt16.bound({ min: -32767}))),
-       
+
        /** Indicates the maximum value of MeasuredValue that can be measured. */
        maxMeasuredValue: Attribute(0x2, TlvNullable(TlvInt16)),
 
        /** Indicates the magnitude of the possible error that is associated with ScaledValue */
-       tolerance: OptionalAttribute(0x3, TlvUInt16.bound({ min: 0, max: 2048 }), { default: 0 }),
+       tolerance: OptionalAttribute(0x3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ }), { default: 0 }),
```
mfucci, line 36: min is already 0 since this is an unsigned int

src/matter/cluster/PressureMeasurementCluster.ts
```diff
@@ -45,7 +45,7 @@ export const PressureMeasurementCluster = Cluster({
        maxScaledValue: OptionalAttribute(0x12, TlvNullable(TlvInt16), { default: 0 }),
 
        /** Indicates the magnitude of the possible error that is associated with ScaledValue */
-       scaledTolerance: OptionalAttribute(0x13, TlvUInt16.bound({ min: 0, max: 2048 }), { default: 0 }),
+       scaledTolerance: OptionalAttribute(0x13, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ }), { default: 0 }),
```
mfucci, line 48: min is already 0 since this is an unsigned int

src/matter/cluster/IlluminanceMeasurementCluster.ts
```diff
@@ -21,14 +21,14 @@ export const IlluminanceMeasurementCluster = Cluster({
...
+
        /** Indicates the minimum value of MeasuredValue that can be measured. */
        minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({ min: 1 }))),
-       
+
        /** Indicates the maximum value of MeasuredValue that can be measured. */
        maxMeasuredValue: Attribute(2, TlvNullable(TlvUInt16.bound({ max: 65534 }))),
-       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 })),
-       
+       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ })),
```
mfucci, line 30: min is already 0 since this is an unsigned int

src/matter/cluster/FlowMeasurementCluster.ts
```diff
@@ -21,13 +21,13 @@ export const FlowMeasurementCluster = Cluster({
...
+
        /** Indicates the minimum value of MeasuredValue that can be measured.  */
        minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({ min: 0 }))),
-       
+
        /** Indicates the maximum value of MeasuredValue that can be measured.  */
        maxMeasuredValue: Attribute(2, TlvNullable(TlvUInt16.bound({ max: 65534 }))),
 
-       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 }), { default: 0 }),
+       tolerance: OptionalAttribute(3, TlvUInt16.bound({ min: 0, max: 2048 /* 0x0800 */ }), { default: 0 }),
```
mfucci, line 31: min is already 0 since this is an unsigned int

src/matter/cluster/FlowMeasurementCluster.ts
```diff
@@ -21,13 +21,13 @@ export const FlowMeasurementCluster = Cluster({
     attributes: {
        /** MeasuredValue represents the flow in m3 /h as follows: MeasuredValue = 10 x Flow  */
        measuredValue: Attribute(0, TlvNullable(TlvUInt16), { default: null }),
-       
+
        /** Indicates the minimum value of MeasuredValue that can be measured.  */
        minMeasuredValue: Attribute(1, TlvNullable(TlvUInt16.bound({ min: 0 }))),
```
mfucci, line 26: min is already 0 since this is an unsigned int

mfucci (changes requested):

src/matter/cluster/OnOffCluster.ts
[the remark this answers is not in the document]
Apollon77: We have that already in other clusters the same way that we define things options if they are dependent on a featureflag. So i would do it the same here so that it can be used if the dev knows this dependency - until we have a better way. So "semantic optional" ;-)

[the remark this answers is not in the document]
Apollon77: I added comments for the "semantic optional" approach in the two relevant clusters till now and also a comment.

@mfucci

For a clean splitup solution we need to either:
* enhance "BitFlag" used in the features to take a value to make is static ... downside would be that the feature struct is then duplicated, OR
* add a "featureDefaults" struct to the Cluster definition to allow to have the "feature definition" to be the same, but then set the feature flag directly. But in this case it should not be required anymore when instanziating the ClusterServer

I would propose to do that after code merge.

I we really do not come to an agreement I will comment the stuff out again and add an Todo - but this limits other Devs to use the fields and commands at all.

Apollon77:
Content included in https://github.com/project-chip/matter.js/pull/59 or as followup

PR #271, opened 2023-03-12
Add linter as last step before code move

Apollon77:
Settings are the same as in matter.js PR that we justed merged


Apollon77:
For the two places where I disabled the @typescript-eslint/require-await check for one line we want the method to be that way. i did not found a better way

mfucci (approved):

PR #273, opened 2023-03-15
Repo update for move

Apollon77:
This is a short readme update to inform user base about the move ...


mfucci (changes requested):
Can you also delete all the code in this MR so there is no confusion?

Apollon77:
I would not like to do this because this is stll currently the way to use it. I would remove the code (as written in the readme text) as soon as we published the first matter-node.js version to npm ...

README.md
```diff
@@ -2,6 +2,18 @@
 
 [![license](https://img.shields.io/badge/license-Apache2-green.svg?style=flat)](https://raw.githubusercontent.com/mfucci/node-matter/master/LICENSE) 
 
+## THIS REPOSITORY HAS BEEN MOVED!!
```
Apollon77, line 5: @JimBuzbee @vves @javierbarellano4v @jdesai61 @Hedda @lboue @dberardo @andyearnshaw @lboue @Flavien FYI, as contributors to code and discussions please feel informed about what will happen here soon ... But in fact we just join forces with project chip to make it even better :-)

mfucci:
Let's avoid to have the same code in two places: let's merge this MR as soon as the code move MR in matter.js is merged and let's have this MR also deleting all the code

Apollon77:
PR updtaed, code removed too

README.md
jdesai61: So node-matter will be a project under packages/  in matter.js?

mfucci: Yes, and we will slowly move down to packages/matter.js all the code which is not node.js specific.

Apollon77: @jdesai61 Yes, I just merged it. it will be in packages/matter-node.js and also the new package when we first publish will be names @project-chip/matter-node.js

mfucci (approved):

Apollon77:
I will do one final matter.js readme update and then I would merge them together.

